PR #31, opened 2023-01-27
test: basic route coverage

jsstevenson:
This provides very basic testing of most of the exposed routes, focused on the basic Dict storage backend. This means that only stubs are provided for testing features that rely on the postgres backend -- I think it'd be better to circle back after we think more about storage and the API library.

I borrowed heavily from @andreasprlic's previous PR, but started from a new branch because a) that PR had been sitting around for a while and b) I wanted us to stick with pytest for now (I think it scales with greater app complexity in a smoother manner than `unittest` but if anyone feels strongly otherwise I'm happy to chat). 

I did have to make some fairly minor rearrangements to the app structure to make it a bit more testable. I think there's more to be done to make testing work in CI (eg we need to be able to perform dependency injection with a mock seqrepo object) but this might be a good start.

There's a lot more work to be done (in addition to the above, we should also make use of fixtures to enable each test module to run independently -- currently some rely on data entered in `test_variation`) but this is a start, hopefully.


Makefile
```diff
@@ -66,13 +72,7 @@ install-extras:
...
 .PHONY: test
 test:
-	python setup.py pytest
-
-#=> tox: execute tests via tox
-.PHONY: tox
-tox:
-	tox
-
+	python -m pytest tests
```
korikuzma, line 75: I can't comment on the exact line, but this
https://github.com/biocommons/anyvar/blob/ec9fbf1c794bb138edae1f456e5621dd0ca13d7b/Makefile#L47

should be changed to 

```
pip install -e .[dev,test]
```

Or we can change setup.cfg to combine both `dev` and `test` into `dev`

src/anyvar/restapi/utils.py
```diff
@@ -1,16 +1,43 @@
...
 
-    The resolver used by connexion doesn't handle relative paths in
-    files. This function replaces all $ref values that have a
-    corresponding entry in ref_map.
+_logger = logging.getLogger("anyvar_api")
 
-      $ref: "file:///vr.json#/definitions/Allele"
 
-    """
+def generate_openapi_yaml():
```
korikuzma, line 15: Add return type

tests/conftest.py
```diff
@@ -1,11 +1,64 @@
...
 
-from anyvar.app import create_app
+from anyvar.restapi.webapp import create_app
+
+
+def pytest_collection_modifyitems(items):
+    """Modify test items in place to ensure test modules run in a given order.
+    When creating new test modules, be sure to add them here.
+    """
+    MODULE_ORDER = [
```
korikuzma, line 15: this can be done in a separate issue if you want, but it might be good to ensure that `len(MODULE_ORDER)` == # of files in tests directory that start with tests_

korikuzma (changes requested):

korikuzma (approved):

Issue #24, opened 2023-01-26
Modularize translator/data proxy layer


jsstevenson:
Pending PR implements a REST-based translator proxy that routes through the variation normalizer's `/normalize` endpoint before storing. Unfortunately, this can't easily replace the necessary data sources for the anyvar sequence/sequence metadata endpoints, so we have to provide SeqRepo separately (this is a little inelegant).

- [ ] It'd be great to get all data from one place somehow. Maybe that means tweaking the normalizer endpoints a bit to expose more SeqRepo stuff. Alternatively, it's not clear to me if the sequence endpoints are a must-have in the long term. Kori has been thinking about this already: https://github.com/GenomicMedLab/cool-seq-tool/issues/96
- [ ] We should be able to inject our own Variation Normalizer instances directly within the process rather than hitting a REST endpoint.
- [ ] We should be able to use VRS Python (REST or internally) in place of the variation normalizer. In #33 I undid some of the modularization work Reece had built, it'd be good to get that back.

PR #33, opened 2023-02-09
feat: use Variation Normalizer for translation

### LINKED ISSUE #29: Allele PUT response should include location ID
Otherwise, the `/locations/` endpoint is a little hard to use

jsstevenson:
I went back and forth on a lot of things here but ultimately I think I want us to move over to FastAPI before I mess with the architecture any more. On the bright side, I think this resolves #29.

* Currently just accessing the normalizer via REST. I don't know if it makes sense to try to run them both together in one Python process. It sort of depends on how much Dockerizing we want to do and how we want to do it. I'll keep #24 open for now.
* I don't know how essential the sequence stuff is, but it unfortunately requires a separate seqrepo dataproxy because the variation normalizer doesn't quite expose the same stuff under `translate_identifier`. It'd be nice to just drop it so that we have one less thing opening seqrepo files.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,6 +36,7 @@ dev =
 test =
 	pytest
         pytest-cov
+        pytest-mock
 normalize =
         variation-normalizer
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,8 +36,6 @@ dev =
 test =
 	pytest
         pytest-cov
-normalize =
-        variation-normalizer
 
 
 [aliases]
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,6 +36,8 @@ dev =
 test =
 	pytest
         pytest-cov
+normalize =
+        variation-normalizer
 
 
 [aliases]
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,6 @@ dev =
 test =
 	pytest
         pytest-cov
-        pytest-mock
 normalize =
         variation-normalizer
 
--- a/src/anyvar/restapi/globals.py
+++ b/src/anyvar/restapi/globals.py
@@ -20,7 +20,6 @@ from ..translate import _Translator
 anyvar_db_fn = os.path.expanduser("/tmp/anyvar.db")
 
 
-
 def _get_g(key: str, creator_function: Callable) -> Any:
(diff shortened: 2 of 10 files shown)


korikuzma:
>   I don't know how essential the sequence stuff is, but it unfortunately requires a separate seqrepo dataproxy because > the variation normalizer doesn't quite expose the same stuff under `translate_identifier`. It'd be nice to just drop it so that > we have one less thing opening seqrepo files.

@jsstevenson Plan to address in [this](https://github.com/GenomicMedLab/cool-seq-tool/issues/96) cool-seq-tool

src/anyvar/restapi/globals.py
```diff
@@ -5,40 +5,68 @@
...
+
+    :param uri: location listening for REST requests
+    :return: instantiated Translator instance
+    """
+
+    if not uri:
+        uri = os.environ.get("ANYVAR_VARIATION_NORMALIZER_URI")
+        if not uri:
+            raise TranslatorSetupException(
+                "No Translator object or URI provided."
```
korikuzma, line 53: Translator object does not get passed 
```suggestion
                "No URI provided."
```

src/anyvar/translate/variation_normalizer.py
```diff
@@ -0,0 +1,61 @@
...
+                f"Variation Normalizer REST endpoint returned server error for {var}"
+            )
+        elif resp.status_code != 200:
+            raise TranslatorConnectionException(
+                f"Variation Normalizer REST endpoint returned {resp.status_code} for {var}"
+            )
+
+        resp_json = resp.json()
+        if resp_json.get("warnings"):
+            raise Exception("TODO figure out what exception should get caught")
```
korikuzma, line 56: Is this being addressed later? Might be nice to make a new issue so we don't forget about it if so

korikuzma (approved):

setup.cfg
```diff
@@ -36,6 +36,8 @@ dev =
 test =
 	pytest
         pytest-cov
+normalize =
```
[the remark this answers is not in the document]
jsstevenson: ope meant to remove this

korikuzma (approved):

PR #34, opened 2023-02-10
style: add new linting checks

jsstevenson:
Btw, I'm adding this commit to an ignore file (https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) once it's squashed


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,28 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
```
korikuzma, line 5: Is there a reason why we only want to run CI on the main branch? I think it's nice when pushing to a branch to see if tests pass before I open a PR

korikuzma (approved):

.github/workflows/ci.yaml
jsstevenson: Yeah I think that makes sense

PR #54, opened 2023-09-17
feat: (re-)add VRS-Python translator

### LINKED ISSUE #51: Create a vrs-python translator
Currently there is only a variation-normalizer translator which will continue to support, but would like the ability for users to have a native vrs-python translator only.

jsstevenson:
close #51 

If memory serves, this was originally part of AnyVar, and I tore it out in favor of the VICC Variation Translator. This PR adds it back in as the default translator layer.

The only data dependency is a SeqRepo instance -- I think it makes sense to fall back on the VRS-Python `create_dataproxy` function and use the same URI schema that it prefers, but set it from the `SEQREPO_DATAPROXY_URI` variable and use REST service on localhost port 5000 by default (rather than erroring if it's unset)


src/anyvar/translate/vrs_python.py
```diff
@@ -0,0 +1,65 @@
...
+
+from biocommons.seqrepo import SeqRepo
+from ga4gh.vrs.dataproxy import SeqRepoDataProxy, _SeqRepoDataProxyBase
+from ga4gh.vrs.extras.translator import Translator
+
+from anyvar.utils.types import VrsPythonVariation
+
+from . import _Translator
+
+SEQREPO_ROOT_DIR = environ.get("SEQREPO_ROOT_DIR", "/usr/local/share/seqrepo/latest")
```
korikuzma, line 13: We should add this to the readme

src/anyvar/translate/vrs_python.py
```diff
@@ -0,0 +1,65 @@
...
+        :param seqrepo_proxy: existing SR proxy instance if available. If not given,
+            try to construct a data proxy using local data at the file address
+            defined under the environment variable ``SEQREPO_ROOT_DIR``, or
+            ``/usr/local/share/seqrepo/latest`` by default.
+        """
+        if not seqrepo_proxy:
+            seqrepo_proxy = SeqRepoDataProxy(SeqRepo(SEQREPO_ROOT_DIR))
+        self.tlr = Translator(
+            data_proxy=seqrepo_proxy
+        )
```
korikuzma, line 31: This could fit on one line I think

```suggestion
        self.tlr = Translator(data_proxy=seqrepo_proxy)
```

src/anyvar/translate/vrs_python.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def translate(self, var: str) -> Optional[VrsPythonVariation]:
+        """Translate provided variation text into a normalized VRS object.
+
+        :param var: user-provided string describing or referencing a variation.
+        :returns: VRS variation object if able to normalize
```
korikuzma, line 37: Not sure if normalize is the right word for this since it could also fail other checks like regex

```suggestion
        :returns: VRS variation object if able to translate to VRS object
```

src/anyvar/translate/vrs_python.py
```diff
@@ -0,0 +1,65 @@
...
+        :raises TranslatorConnectionException: if translation request returns error
+        """
+        try:
+            return self.tlr.translate_from(var)
+        except ValueError:
+            raise NotImplementedError(
+                f"{var} isn't supported by AnyVar yet."
```
korikuzma, line 44: I think this is a VRS-Python issue, right? 

```suggestion
                f"{var} isn't supported by VRS-Python yet."
```

korikuzma (changes requested):
Looks good. Just some minor suggestions

korikuzma (approved):

PR #63, opened 2023-09-29
chore: remove unused code

korikuzma:
Remove unused code. We can add things back in the future if needed

Notes:
* Removed clinvarparser, variation normalizer rest translator, digest, formats, + ordered enum
* Removed DEFAULT_TRANSLATE_URI


jsstevenson (dismissed):
👍 

there is... a view of AnyVar where the VICC normalizer is a first-class citizen, so I think we should be sure to maintain the translate interface abstraction (rather than becoming too intertwined with VRS-Python specifically) but I'm happy to not worry too much about the normalizer backend for now

jsstevenson (approved):

Issue #71, opened 2024-01-05
Update put_object to take advantage of vrs-python API changes and avoid unnecessary VRS computation


ehclark:
The changes for https://github.com/ga4gh/vrs-python/issues/305 include a new parameter for `vrs_enref` that returns both the object and its identifier.  This should be called from `AnyVar.put_object` to avoid the subsequent call to `ga4gh_identify`.

PR #73, opened 2024-01-08
Improve performance of AnyVar.put_object() by eliminating unnecessary digest computation

ehclark:
In `AnyVar.put_object()`, use new parameter to `vrs_enref()` method to return both the object and identifier.  This eliminates the need to recompute the already computed digest and improve performance.

See also #71

### DIFF
--- a/src/anyvar/anyvar.py
+++ b/src/anyvar/anyvar.py
@@ -82,7 +82,7 @@ class AnyVar:
         :return: Object digest if successful, None otherwise
         """
         try:
-            id, v = vrs_enref(variation_object, self.object_store, True)
+            id, _ = vrs_enref(variation_object, self.object_store, True)
         except ValueError:
             return None
         return id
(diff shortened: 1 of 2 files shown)


jsstevenson (approved):

src/anyvar/anyvar.py
```diff
@@ -83,10 +82,10 @@ def put_object(self, variation_object: VrsObject) -> Optional[str]:
         :return: Object digest if successful, None otherwise
         """
         try:
-            v = vrs_enref(variation_object, self.object_store)
+            id, v = vrs_enref(variation_object, self.object_store, True)
```
korikuzma, line 85: `v` isn't being used anywhere, so we could just set to `_`

```suggestion
            id, _ = vrs_enref(variation_object, self.object_store, True)
```

korikuzma (approved):

korikuzma (approved):

Issue #65, opened 2023-12-12
Add a parameter to the /vcf endpoint that ensures all async VRS object writes are completed before returning


ehclark:
The Snowflake storage backend writes VRS object batches asynchronously.  Read operations may not reflect unwritten batches.  Add a `flush()` API endpoint that would return when pending batches are written.  Specifically, the flush call would wait for any batches pending at the time the call was made to be completed.

See also #64

PR #75, opened 2024-01-08
Add parameter to /vcf endpoint to force async writes to complete before returning response

ehclark:
The new Snowflake object store writes VRS objects to the database asynchronously to allow better response times.  However, since searches and fetches only read committed database state, this creates the opportunity for reads to not reflect the most recent VRS objects.

The current default behavior, and default behavior moving forward, is for all writes to complete before the response is returned to the client.  To enable writes to complete asynchronously after the response has been returned, clients can add the `allow_async_writes=yes` parameter to the request.

See also #65

### DIFF
--- a/src/anyvar/storage/postgres.py
+++ b/src/anyvar/storage/postgres.py
@@ -127,6 +127,10 @@ class PostgresObjectStore(_Storage):
             cur.execute("DELETE FROM vrs_objects WHERE vrs_id = %s;", [name])
         self.conn.commit()
 
+    def wait_for_writes(self):
+        """Return true once any currently pending database modifications have been completed."""
+        pass
+
     def close(self):
         """Terminate connection if necessary."""
         if self.conn is not None:
--- a/src/anyvar/storage/postgres.py
+++ b/src/anyvar/storage/postgres.py
@@ -128,8 +128,9 @@ class PostgresObjectStore(_Storage):
         self.conn.commit()
 
     def wait_for_writes(self):
-        """Return true once any currently pending database modifications have been completed."""
-        pass
+        """Returns once any currently pending database modifications have been completed.
+        The PostgresObjectStore does not implement async writes, therefore this method is a no-op
+        and present only to maintain compatibility with the `_Storage` base class"""
 
     def close(self):
         """Terminate connection if necessary."""
--- a/src/anyvar/storage/snowflake.py
+++ b/src/anyvar/storage/snowflake.py
@@ -207,7 +207,7 @@ class SnowflakeObjectStore(_Storage):
         if self.batch_thread is not None:
             # short circuit if the queue is empty
             with self.batch_thread.cond:
-                if len(self.batch_thread.pending_batch_list) == 0:
+                if not self.batch_thread.pending_batch_list:
                     return
 
             # queue an empty batch
@@ -216,12 +216,7 @@ class SnowflakeObjectStore(_Storage):
             # wait for the batch to be removed from the pending queue
             while True:
                 with self.batch_thread.cond:
-                    if (
-                        len(
-                            list(filter(lambda x: x is batch, self.batch_thread.pending_batch_list))
-                        )
-                        > 0
-                    ):
+                    if list(filter(lambda x: x is batch, self.batch_thread.pending_batch_list)):
                         self.batch_thread.cond.wait()
                     else:
                         break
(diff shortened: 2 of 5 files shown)


src/anyvar/storage/snowflake.py
```diff
@@ -202,6 +202,30 @@ def __delitem__(self, name: str) -> None:
...
+
+            # queue an empty batch
+            batch = []
+            self.batch_thread.queue_batch(batch)
+            # wait for the batch to be removed from the pending queue
+            while True:
+                with self.batch_thread.cond:
+                    if (
+                        len(
+                            list(filter(lambda x: x is batch, self.batch_thread.pending_batch_list))
```
jsstevenson, line 221: Double checking that `is` is appropriate here -- the object doesn't get copied or moved around while being held on the queue?

jsstevenson (approved):

src/anyvar/storage/postgres.py
```diff
@@ -127,6 +127,10 @@ def __delitem__(self, name: str) -> None:
             cur.execute("DELETE FROM vrs_objects WHERE vrs_id = %s;", [name])
         self.conn.commit()
 
+    def wait_for_writes(self):
+        """Return true once any currently pending database modifications have been completed."""
+        pass
```
korikuzma, line 132: I don't think the pass is necessary. Is there an existing issue to do this for postgres backend? 

```suggestion
```

src/anyvar/storage/snowflake.py
```diff
@@ -202,6 +202,30 @@ def __delitem__(self, name: str) -> None:
             cur.execute(f"DELETE FROM {self.table_name} WHERE vrs_id = ?;", [name])  # nosec B608
         self.conn.commit()
 
+    def wait_for_writes(self):
+        """Return true once any currently pending database modifications have been completed."""
```
korikuzma, line 206: I might be missing something, but I don't think this method returns true

src/anyvar/storage/snowflake.py
```diff
@@ -202,6 +202,30 @@ def __delitem__(self, name: str) -> None:
...
+        if self.batch_thread is not None:
+            # short circuit if the queue is empty
+            with self.batch_thread.cond:
+                if len(self.batch_thread.pending_batch_list) == 0:
```
korikuzma, line 210:
```suggestion
                if not self.batch_thread.pending_batch_list:
```

src/anyvar/storage/snowflake.py
```diff
@@ -202,6 +202,30 @@ def __delitem__(self, name: str) -> None:
...
+            self.batch_thread.queue_batch(batch)
+            # wait for the batch to be removed from the pending queue
+            while True:
+                with self.batch_thread.cond:
+                    if (
+                        len(
+                            list(filter(lambda x: x is batch, self.batch_thread.pending_batch_list))
+                        )
+                        > 0
+                    ):
```
korikuzma, line 224: I think you could also do this:

```suggestion
                    if list(filter(lambda x: x is batch, self.batch_thread.pending_batch_list)):
```

src/anyvar/storage/postgres.py
jsstevenson: IMO we shouldn't worry about implementing it PG-side until there's a use case/demand for doing something like queueing. I think there's a world in which it could be preferable to retain the PG implementation as a more streamlined read/write option.

That said, it might be good to note in the docstring that this method is just a stub to retain compatibility with the base class, but shouldn't do anything unless we decide to implement queueing logic for the postgres writer.

korikuzma: Then I don't think it should be an abstractmethod in the `_Storage` class and should only exist in the snowflake backend

src/anyvar/storage/snowflake.py
ehclark: Yes this should be a strict identity check.  The empty batch is essentially a sentinel object and batch objects are not modified as they move through the queue.

src/anyvar/storage/postgres.py
ehclark: I think the `wait_for_writes()` method should remain on `_Storage` because it is called from `anyvar.restapi.main` which does not understand which backend it is using.  The REST API method implementations only calls methods on the `_Storage` base class.  Without it on the base class, calling `/vcf` with `allow_async_write=no` results in a 500 error if the backend store in Postgres.

I cleaned up the comments and removed the `pass` line.

korikuzma: @ehclark that makes sense. Thanks

korikuzma (approved):

ehclark:
@jsstevenson  Could you do another review when you get a chance.  Thanks!

jsstevenson (approved):
👍 lgtm

Issue #66, opened 2023-12-12
Add a shutdown API to enable graceful termination of AnyVar


ehclark:
The Snowflake storage backend writes VRS object batches asynchronously.  Untimely termination could therefore result in pending batches being discarded.  Providing an API for graceful shutdown would allow for all pending batches to be written prior to termination.  This API should be callable as a PreStop hook in Kubernetes: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/

See also #64

ehclark:
Given the process model used by ASGI, it would be improper to place a shutdown API in the app itself.  Startup/shutdown is the responsibility of the web server, such as uvicorn or gunicorn.

Instead, AnyVar app should perform a graceful shutdown of the storage integration when a shutdown is occurs.  This can be implemented by [lifespan events](https://fastapi.tiangolo.com/advanced/events/), which are already in use for startup.

PR #76, opened 2024-01-11
Cleanup storage integration and write pending batches on shutdown

ehclark:
The integration used by AnyVar to store, retrieve and query VRS objects should be provided an opportunity to complete pending work, shutdown and clean up when an AnyVar worker is shutdown.  The FastAPI provides lifespan events that can be used for this purpose.

The changes simply make the `close()` method part of the `_Storage` API definition and then call the method when the shutdown occurs.  Implementations of the `close()` already existed for both the PostgreSQL and Snowflake storage implementations.

The existing startup event handler was moved to a new lifespan method and the call to `close()` added to the lifespan method.  Event handlers have been deprecated in favor of [lifespan](https://fastapi.tiangolo.com/advanced/events/).

See also #66

### DIFF
--- a/src/anyvar/restapi/main.py
+++ b/src/anyvar/restapi/main.py
@@ -1,6 +1,7 @@
 """Provide core route definitions for REST service."""
 import logging
 import tempfile
+from contextlib import asynccontextmanager
 from http import HTTPStatus
 
 import ga4gh.vrs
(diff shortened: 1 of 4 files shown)


jsstevenson (approved):
👍 this is great, I didn't realize they'd deprecated the startup event

korikuzma (approved):

ahwagner (approved):
This is great. TIL about FastAPI lifespans! Thanks for this addition @ehclark.

Issue #91, opened 2024-04-18
Add back missing response models to openapi.json


ehclark:
Split off from https://github.com/biocommons/anyvar/issues/85

Same underlying problem as https://github.com/cancervariants/variation-normalization/issues/543

Changed the response models to return `Dict` instead of the types from `ga4gh.vrs.models`.  This allows the full, dereferenced VRS object to be returned from the API endpoints instead of the version produced by the serialization logic embedded in the model hierarchy.

jsstevenson:
Closing in on a resolution in https://github.com/ga4gh/vrs-python/pull/435

PR #92, opened 2024-04-18
Update unit tests to match the new VRS 2.0 digests

ehclark:
Updated the unit tests inputs and expected outputs to match VRS 2.0.

In addition, the following changes were made:
- `__getitem__` now throws KeyError on missing key to match API expectations for `MutableMapping`
- Response models no longer reference `ga4gh.vrs.models` classes (see #91)
- `/locations/{id}` and `/variation/{id}` behave consistently and return 404 when `id` is not found


korikuzma (approved):

jsstevenson (approved):
👍

PR #94, opened 2024-04-26
cicd: run pytest in github actions

jsstevenson:
Was curious to see if this works (or at least ~works)


jsstevenson:
(seqrepo rest service still needs some local seqrepo data to be available as a mounted volume so this doesn't quite work)

PR #95, opened 2024-07-19
fix: update ga4gh.vrs version + add back VRS type anno in models

### LINKED ISSUE #91: Add back missing response models to openapi.json

korikuzma:
close #91


jsstevenson (approved):
I think this fix also lets us revert the `model_dump` calls in the API responses?

```python
    if location:
        return {"location": location.model_dump(exclude_none=True)}
```

korikuzma:
I removed my anyvar container, so I didn't test the last commit. Will come back to this later to test and request review afterwards.

ehclark (approved):
Added an additional change to the response schema to get the unit tests to pass.  The responses for the register endpoints should return a model object if the call to model_dump() is removed.

korikuzma:
@ehclark thank you!

Issue #100, opened 2024-10-21
Snowflake storage connector not working because of change to SnowflakeBatchAddMode enum


ehclark:
**Describe the bug**
The `SnowflakeBatchAddMode` enum was changed from a `StrEnum` to an `Enum` for compatibility with Python 3.10.  But the code in the Snowflake connector that processes was not properly updated. 

**To Reproduce**
1. Clone AnyVar repo
2. Prepare for development: `make devready`
3. Install optional deps: `pip install -e .[dev,tests,snowflake]`
4. Run Snowflake unit tests: `python -m pytest tests/storage/test_snowflake.py`

All tests fail with error: Exception: batch_add_mode must be one of 'merge', 'insert_notin', or 'insert'

**Expected behavior**
Unit tests should pass

Issue #99, opened 2024-10-21
AnyVar startup fails due to /vcf response model validation error


ehclark:
**Describe the bug**
The `/vcf` endpoint method returns either a `FileResponse` (when successful) or a `dict` (when not successful).  `FileResponse` is not a valid Pydantic field type.

**To Reproduce**
1. Clone AnyVar repo
2. Prepare for development: `make devready`
3. Start AnyVar: `uvicorn anyvar.restapi.main:app`
fastapi.exceptions.FastAPIError: Invalid args for response field! Hint: check that starlette.responses.FileResponse | dict is a valid Pydantic field type. If you are using a return type annotation that is not a valid Pydantic field (e.g. Union[Response, dict, None]) you can disable generating the response model from the type annotation with the path operation decorator parameter response_model=None. Read more: https://fastapi.tiangolo.com/tutorial/response-model/

**Expected behavior**
Startup is successful

Issue #98, opened 2024-10-21
sqlalchemy missing from required dependencies list


ehclark:
The sqlalchemy library is required for both Postgres and Snowflake storage backends.  It should be in the list of required dependencies.

**To Reproduce**
1. Clone anyvar repo
2. Follow the instructions to prepare for development: `make devready`
3. Start anyvar: `uvicorn anyvar.restapi.main:app`
ERROR: ModuleNotFoundError: No module named 'sqlalchemy'

**Expected behavior**
AnyVar starts up successfully

PR #101, opened 2024-10-21
Fix problems introduced by project configuration changes and dependency upgrades

ehclark:
The latest round of updates to the project configuration and dependency changes caused some things to break.
- Added sqlalchemy to list of required dependencies (#98)
- Disable FastAPI validation of response model for `/vcf` endpoint since the `FileResponse` is not valid pydantic field (#99)
- Fix handling of batch add mode setting in Snowflake connector (#100)
- Change variant test data to use "EFO:" instead of "efo:" in copy number objects because Pydantic deserialization was failing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "python-multipart",  # required for fastapi file uploads
     "uvicorn",
     "ga4gh.vrs[extras]~=2.0.0a10",
+    "sqlalchemy~=1.4.54",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "fastapi>=0.95.0",
     "python-multipart",  # required for fastapi file uploads
     "uvicorn",
-    "ga4gh.vrs[extras]~=2.0.0a10",
+    "ga4gh.vrs[extras]~=2.0.0a11",
     "sqlalchemy~=1.4.54",
 ]
 dynamic = ["version"]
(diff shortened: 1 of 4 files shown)


jsstevenson (approved):
👍 whoop, sorry about that. Thanks for the check-up.

pyproject.toml
```diff
@@ -19,7 +19,8 @@ dependencies = [
     "fastapi>=0.95.0",
     "python-multipart",  # required for fastapi file uploads
     "uvicorn",
-    "ga4gh.vrs[extras]~=2.0.0a10",
+    "ga4gh.vrs[extras]~=2.0.0a11",
+    "sqlalchemy~=1.4.54",
```
korikuzma, line 23: I'm not sure if we needed to add `sqlalchemy` to required dependencies (since not everyone will be using sqlalchemy). I think the optional `snowflake` dep handles this?

jsstevenson: @korikuzma I had the same thought but I think it's probably fine? (really there is only one user right now)

Either way, since it's named as an imported module it should be declared as an explicit dependency even if other dependencies will probably include it (in the sql groups or in main)

korikuzma: Right, I think it would be nice to keep this in the `snowflake` optional dep to keep things lightweight

korikuzma: It doesn't have to be now, but I'm at least going to make an issue to address in the future

korikuzma: Commented in issue -- I assumed we were only using psycopg2 for postgres. @jsstevenson pointed out it's also used in snowflake. This makes sense now

PR #108, opened 2024-11-18
VCF annotation via asynchronous request-response pattern

### LINKED ISSUE #107: Implement asynchronous request-reply for VCF annotation
**Is your feature request related to a problem? Please describe.**
Long running requests do not play nicely with cloud-based API hosting services.  In general, API response times are expected to be less than 30 seconds.  Long running requests unnecessarily consume network resources.

**Describe the solution you'd like**
Implement an asynchronous request-reply pattern for annotation of a VCF file with VRS IDs using the [Celery distributed task queue](https://docs.celeryq.dev/en/stable/index.html) with Redis as a backend and broker.  While the asynchronous request-reply pattern can be implemented in slightly different ways, I intend to follow the [semantics defined by Snowflake](https://docs.snowflake.com/en/sql-reference/external-functions-implementation#asynchronous-vs-synchronous-remote-services) for integration purposes.

AnyVar would remain deployable and functional without Redis and other async dependencies.

ehclark:
See also #107

The annotation of a large VCF file with VRS IDs can run for a long time.  A synchronous request-response pattern where the client is waiting for a response can lead to instability and dropped connections in many environments.  For long running processes, a better option is an asynchronous request-response pattern.  The client submits a request and receives a 202 response and polling location.  The client polls at some interval, receiving a 202 if the request is still pending, a 200 and the result if the request is completed, or a 500 if the request failed.

AnyVar uses the Celery distributed task queue for task management and result caching.  The asynchronous feature is optional and the changes for this PR are backwards compatible.

See the [README-async.md](README-async.md) file for detailed documentation of the async feature and how to enable it.

Closes #107

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,8 @@ test = [
     "pytest",
     "pytest-cov",
     "pytest-mock",
-    "httpx"
+    "httpx",
+    "celery[pytest]",
 ]
 dev = [
     "ruff==0.5.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
     "uvicorn",
     "ga4gh.vrs[extras]~=2.0.0a11",
     "sqlalchemy~=1.4.54",
-    "pyaml",
+    "pyyaml",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ queueing = [
     "celery[redis]~=5.4.0",
     "aiofiles",
 ]
-tests = [
+test = [
     "pytest",
     "pytest-cov",
     "pytest-mock",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,12 +21,17 @@ dependencies = [
     "uvicorn",
     "ga4gh.vrs[extras]~=2.0.0a11",
     "sqlalchemy~=1.4.54",
+    "pyaml",
 ]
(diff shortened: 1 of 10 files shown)


ehclark:
Thanks @korikuzma.  No rush.  I am going to need to make some additional tweaks anyway and will be out next week.  I have reverted it to a draft.  But if you want to take a quick look next week just to make sure it looks directionally correct that would be helpful.

pyproject.toml
```diff
@@ -21,17 +21,23 @@ dependencies = [
     "uvicorn",
     "ga4gh.vrs[extras]~=2.0.0a11",
     "sqlalchemy~=1.4.54",
+    "pyaml",
```
jsstevenson, line 24: out of curiosity, any reason for `pyaml` and not `pyyaml`?

korikuzma (dismissed):
Sorry for delay. Did quick review and seems good to me!

pyproject.toml
ehclark: Probably the code snippet I used to import logging config used `pyaml`.  But it appears that `pyyaml` would be sufficient.  So that can be changed.

korikuzma (approved):

ehclark:
@korikuzma @jsstevenson @theferrit32 

I am going to go ahead and merge this on Monday if there are no other comments/reviews.  If you want or need more time to look at it let me know.

jsstevenson:
👍 looks good to me

Issue #109, opened 2024-12-09
Suppress Pydantic warning messages


ehclark:
When annotating a VCF file with VRS IDs, Pydantic emits the following message for each allele:
```
/anyvar/venv/3.11/lib/python3.11/site-packages/pydantic/main.py:426: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue: Unexpected Value
  PydanticSerializationUnexpectedValue: Expected `SequenceLocation` but got `str` with value `'ga4gh:SL.Uyy0TC8kQG7SgZBvd_rYuZ7kCwb8M_tC'` - serialized value may not be as expected
  return self.__pydantic_serializer__.to_python(
```

This message is logged via `print()` so there is no way to suppress it via log configuration.  In a high volume deployment, this creates copious amounts of log noise.

The underlying cause would appear to be that `AnyVar.put_object` encodes references, replacing the `location` attribute with a `str`.  Then when the VRS object is serialized for database insertion, the mismatch occurs and results in the warning message.

The only apparent way to suppress warning messages is to add `warnings=False` to calls to the `model_dump()` method.

jsstevenson:
This also seems like a couple of upstream problems --

* serialization issues should be logged, not printed
* this should not be causing a serialization issue

jsstevenson:
Oh, is Pydantic doing this via the stdlib `warnings` module? I think there are ways to restrict how library warnings are emitted, a la

```python
import warnings
warnings.filterwarnings("ignore", module="pydantic")
```

ehclark:
Thanks @jsstevenson you are correct about the stdlib here.  Adding that snippet does indeed suppress the output and I have added that suppression as the default (overridable by an env var).

But is there another issue we need to look at with it happening in the first place?  It does not seem harmful from an AnyVar standpoint.

jsstevenson:
I think it's probably benign in impact, but indicative that something in VRS-Python wasn't quite written correctly... from the schema's perspective, the serializer should be expecting a string (IRI) as a possibility for that property. I may or may not look this up on my own time but you shouldn't feel the need to keep this issue open on my account.

PR #110, opened 2024-12-09
Suppress Pydantic warnings by default

ehclark:
Pydantic uses the stdlib `warnings` module to output serialization warnings.  When annotating VCF files at scale, a warning is produced for each allele, which results in very noisy output.

This change causes the Pydantic warnings to be suppressed by default.  The can be re-enabled by setting the `ANYVAR_SHOW_PYDANTIC_WARNINGS` environment variable to any value.

See #109

### DIFF
--- a/src/anyvar/anyvar.py
+++ b/src/anyvar/anyvar.py
@@ -8,6 +8,7 @@ import logging
 import logging.config
 import os
 import pathlib
+import warnings
 from collections.abc import MutableMapping
 from urllib.parse import urlparse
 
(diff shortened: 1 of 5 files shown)


src/anyvar/anyvar.py
```diff
@@ -19,6 +20,10 @@
 from anyvar.translate.vrs_python import VrsPythonTranslator
 from anyvar.utils.types import VrsObject
 
+# Suppress pydantic warnings unless otherwise indicated
+if os.environ.get("ANYVAR_SHOW_PYDANTIC_WARNINGS", None) is None:
```
jsstevenson, line 24: I kind of thought there was a linting rule about not supplying `None` as an arg to `get()` since it's the default but I guess not?

jsstevenson (approved):
👍

PR #113, opened 2024-12-10
build: use importlib (pkg_resources is deprecated)

jsstevenson:
`pkg_resources` is no longer available by default in python 3.12: https://github.com/python/cpython/issues/95299

this PR uses the recommend `importlib` replacement. It also runs the Snowflake unit tests in CI as a base confirmation that the library is importable, since these tests are sufficiently mocked out.


jsstevenson:
...the corresponding fix in `hgvs` is still unreleased...

PR #115, opened 2025-02-14
refactor: use latest VRS-Python release for Anyvar

jsstevenson:
Draft based on https://github.com/ga4gh/vrs-python/pull/502. 

This isn't an API breaking change insofar as the end user shouldn't see a difference, but the public API of the modules themselves do change a bit.


src/anyvar/extras/vcf.py
```diff
@@ -1,117 +1,77 @@
...
-        require_validation: bool = True,  # noqa: ARG002
-    ) -> None:
-        """Get VRS Object given `vcf_coords`. `vrs_data` and `vrs_field_data` will
-        be mutated. Generally, we expect AnyVar to use the output_vcf option rather than
-        the pickle file.
+        vcf_coords: str,  # noqa: ARG002
+        vrs_allele: Allele,
+        **kwargs,
+    ) -> Allele | None:
+        av = kwargs.get("av")
```
ehclark, line 66: Is there a reason that `self.av` will not work here to get a reference to the AnyVar instance?

jsstevenson:
this will need a few small updates again, pending the latest vrs-python release

src/anyvar/extras/vcf.py
jsstevenson: (forgot to respond) -- I think this is an artifact of a refactor that I did mid-PR. Will double-check

Krt-11:
@jennifer-bowser and I worked on this issue in a different PR (#159) but we ran into some problems. After pair programming with @theferrit32 we figured out that our changes from that PR were almost ready and needed some minor tweaks.

Hence Jennifer closed that issue and we decided to move forward with James changes and added the remaining changes from Issue #159.

theferrit32 (approved):

PR #153, opened 2025-05-12
Improve Testing Process

### LINKED ISSUE #152: Improve testing process (README + Makefile)
### Feature description

The setup required for running tests has changed a bit since the README was written. It would be good to update the docs to make it easier for devs to run tests.

Also, tests require having all extra dependencies installed. It would be nice to include this automatically in the `make tests` command.

Additionally, it would be nice to have a small SQL script that mirrors `postgres_init.sql` to setup the DB required for tests.

Also, the `conftest.py` file needs to be updated to correct the `storage_uri` var (see "Implementation Details" below)

### Use case

As a developer, I want to be able to run our tests easily, so I can ensure I'm writing quality code.

### Acceptance Criteria

Developers can follow the steps outlined in the README, without requiring any other outside knowledge or processes, to run the repo's test suite.

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Implementation details

# Makefile
Add the following line to the `test` section of the makefile:
```shell
pip install -e '.[postgres,snowflake,queueing,test,dev]’
```
This would go before the `python -m pytest tests` line.

# Test DB Setup SQL Script
I would suggest copying `postgres_init.sql` and naming the copy something like `postgres_test_init.sql`, then changing all the instances where the script refers to the `anyvar` database to be `anyvar_test`, since this is the [database name the tests expect](https://github.com/biocommons/anyvar?tab=readme-ov-file#testing) according to the README.

The file should look like this:
```SQL
DROP DATABASE IF EXISTS anyvar_test;
DROP USER IF EXISTS anyvar;

CREATE USER anyvar WITH PASSWORD 'anyvar-pw';
CREATE DATABASE anyvar_test WITH OWNER anyvar;
```

# README
- Tests require you to have a postgres instance running. Adding instructions on how to do this to the README would be helpful.

- The README outlines how to install SeqRepo locally, but the tests rely on a SeqRepo REST service. It would be nice to tell devs how to optionally point the tests at their local SeqRepo

# `conftest.py`
In `tests/conftest.py` > `storage()`, the following line is incorrect:
```python
else:
    storage_uri = "postgresql://postgres:postgres@localhost:5432/anyvar_test"
```
It needs to be updated to:
```python
else:
    storage_uri = "postgresql://postgres:postgres@localhost:5432/anyvar_test"
```

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

None

jennifer-bowser:
closes #152


Makefile
```diff
@@ -73,6 +73,7 @@ cqa:
 #=> test: execute tests
 .PHONY: test
 test:
+	pip install -e '.[postgres,snowflake,queueing,test,dev]'
```
korikuzma, line 76: I'm not sure we want to install the deps each time... maybe consider adding a make target for installing certain optional deps and updating the documentation to state that test deps should already be installed and inside the virtual env before running?

jennifer-bowser: That is a much better idea! I'll do that

src/anyvar/storage/postgres_test_init.sql
```diff
@@ -0,0 +1,5 @@
+DROP DATABASE IF EXISTS anyvar_test;
+DROP USER IF EXISTS anyvar;
+
+CREATE USER anyvar WITH PASSWORD 'anyvar-pw';
+CREATE DATABASE anyvar_test WITH OWNER anyvar;
```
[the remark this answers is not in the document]
jennifer-bowser: Oh wow, I am blind 🤦🏻‍♀️ Yes let's use this one!

theferrit32 (approved):

Krt-11 (dismissed):

Makefile
```diff
@@ -70,6 +70,10 @@ cqa:
 	black --check src
 	bandit -ll -r src
 
+.PHONY: testready
+testready:
+	pip install -e '.[postgres,snowflake,queueing,test,dev]'
```
korikuzma, line 75: Do tests need the `dev` dependency?

jennifer-bowser: Oh good point, I guess they don't! I'll remove that

Issue #137, opened 2025-04-16
Add Docker setup instructions


korikuzma:
### Feature description

@theferrit32 made a great [Getting Started doc](https://docs.google.com/document/d/1YIcoI1YlO6okC9i3kixGHzt-AH2uvWBGj1SUm3bhpHg/edit?usp=sharing) for the GA4GH GKS Pre-Connect Hackathon. I don't think this documentation exists anywhere in the repo, but it should be added.

### Use case

* I want to be able to easily spin up AnyVar using Docker
* We're going to be ramping up AnyVar effort and this would help make onboarding new devs to this project smoother

### Acceptance Criteria

This would be an ice sub-issue for #131

### Contribution

None

theferrit32:
Add `docs/` directory to the repository and move the `README-async` to it and also add a 'getting started' markdown file with key points/summarized steps from the google doc linked above

korikuzma:
@jennifer-bowser @Krt-11 This would be a great issue for you two to work on as well

PR #162, opened 2025-05-21
feat: set environment variables via a `.env` file (issue 158)

### LINKED ISSUE #158: Add env.example to document environment variables
Include both first part variables and variables which can configure dependencies like seqrepo/uta.

The application should also automatically load a `.env` file, shouldn't need to source + export

jennifer-bowser:
Closes #158


pyproject.toml
```diff
@@ -134,7 +135,9 @@ select = [
     "PERF",  # https://docs.astral.sh/ruff/rules/#perflint-perf
     "FURB",  # https://docs.astral.sh/ruff/rules/#refurb-furb
     "RUF",  # https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf
-    "FAST", # https://docs.astral.sh/ruff/rules/#fastapi-fast
+    "FAST001", #https://docs.astral.sh/ruff/rules/#fastapi-fast
+    "FAST002", #https://docs.astral.sh/ruff/rules/#fastapi-fast
+    "FAST003" #https://docs.astral.sh/ruff/rules/#fastapi-fast
```
jennifer-bowser, line 140: I couldn't commit without updating this, as I was getting errors about `FAST` not being a recognized rule. I replaced the broken code with the ones indicated here [https://docs.astral.sh/ruff/rules/#fastapi-fast](https://docs.astral.sh/ruff/rules/#fastapi-fast).

jennifer-bowser:
TODO: I need to update the README quick start guide to use a .env file instead of exporting environment variables; but I need to wait for Issue https://github.com/biocommons/anyvar/issues/137 to get merged first because it implements a major documentation overhaul, and trying to edit the README now will just result in major merge conflicts once that's released.

pyproject.toml
```diff
@@ -43,7 +44,7 @@ test = [
     "celery[pytest]",
 ]
 dev = [
-    "ruff==0.11.2",
```
jennifer-bowser, line 46: I upgraded ruff because I was getting errors about `FAST` being an "unknown rule selector". I'm not sure I fixed the issue, but I figured it can't hurt to keep the upgrade.

korikuzma (commented):
Something to consider looking into: https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#env-file

jennifer-bowser:
> Something to consider looking into: https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#env-file

@korikuzma are you suggesting I replace values in the compose.yaml file with environment variables? I like that idea! It'd definitely cut down on repetition.

jsstevenson:
A few side thoughts: 

* I think this PR should be classified as a `feat`, not a `chore`, since it makes changes to runtime code/the user experience
* I've been meaning to look into [pydantic-settings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#validation-of-default-values), it's sort of like a superset of `dotenv` features with some added validation, and has you define configs in a central place (not sure if that's a good thing or a bad thing, esp when configs span across libraries). I don't know if it's worth serious consideration yet but maybe we can revisit in the future

.env.example
```diff
@@ -0,0 +1,46 @@
+############
+# REQUIRED #
+############
+UTA_DB_URL=postgresql://anonymous@localhost:5432/uta/uta_20241220			# See `docs/uta.md` for more info
+SEQREPO_DATAPROXY_URI=seqrepo+file:///usr/local/share/seqrepo/2024-12-20	# See `docs/seqrepo` for more info
+ANYVAR_STORAGE_URI=postgresql://anyvar:anyvar-pw@localhost:5432/anyvar		# Set to `null` if running AnyVar without storage
```
jsstevenson, line 6: based on https://github.com/biocommons/anyvar/blob/964e7c274047011e343e2c9d477e518a474fc29b/src/anyvar/anyvar.py#L66

I think this actually needs to be an empty string, not the string "null", to run w/o storage. Needs to be fixed in a few other places too.

README.md
```diff
@@ -19,17 +19,24 @@ identifiers, and register biological sequence variation according to the
...
 	```
-2. If desired, create/start a database for AnyVar to use. See [Optional Dependencies - Databases](#optional-dependencies---databases) for detailed instructions on how to set up a database. Otherwise, set the `ANYVAR_STORAGE_URI` environment variable to `null` to use AnyVar without a database.
+
+
+2. **Set Environment Variables**
+
+    Create a new file in the root directory of the repo called `.env`. In the next two steps, you will populate this file. You may want to see `.env.example` for reference.
+
+
+2. If desired, create/start a database for AnyVar to use. See [Optional Dependencies - Databases](#optional-dependencies---databases) for detailed instructions on how to set up a database. Otherwise, set the `ANYVAR_STORAGE_URI` environment variable to `null` in your `.env` file to use AnyVar without a database.
```
jsstevenson, line 35: for what it's worth, this and subsequent numbers should be incremented since you added a new step, although the GitHub markdown viewer will still render it correctly anyway (TIL)

```suggestion
3. If desired, create/start a database for AnyVar to use. See [Optional Dependencies - Databases](#optional-dependencies---databases) for detailed instructions on how to set up a database. Otherwise, set the `ANYVAR_STORAGE_URI` environment variable to `null` in your `.env` file to use AnyVar without a database.
```

jsstevenson (changes requested):

README.md
jennifer-bowser: Oh lawdy I can't believe I forgot to fix the numbering. Thank you, good catch!

theferrit32 (approved):

PR #181, opened 2025-06-25
feat: resolve warnings, annotate routes with response model

jsstevenson:
Mostly quality of life/readability upgrades, but does have runtime implications insofar as it protects against future deprectations

* Resolve a few warnings about deprecated things -- config class -> ConfigDict, example -> examples, schema_extra -> json_schema_extra, dict() -> model_dump(). Suppress a sqlalchemy warning because we do the thing it recommends.
* Also clean up request/response model docstrings to specifically identify routes
* Annotate routes with response models as return objects rather than using the `response_model` param (enables IDE type checking)

### DIFF
--- a/src/anyvar/restapi/schema.py
+++ b/src/anyvar/restapi/schema.py
@@ -1,7 +1,7 @@
 """Provide response definitions to REST API endpoint."""
 
 from enum import Enum
-from typing import Any, Optional
+from typing import Optional
 
 from ga4gh.vrs import models
 from pydantic import BaseModel, ConfigDict, StrictInt, StrictStr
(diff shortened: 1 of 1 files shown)


src/anyvar/restapi/schema.py
```diff
@@ -26,38 +26,31 @@ class DependencyInfo(BaseModel):
...
 class InfoResponse(BaseModel):
-    """Describe response for the /info endpoint"""
+    """Describe response for the GET /info endpoint"""
+
+    model_config = ConfigDict(
+        json_schema_extra={
+            "examples": [
+                {
+                    "anyvar": {"version": "0.1.2.dev58+g81eb592.d20230316"},
+                    "ga4gh_vrs": {"version": "0.7.6"},
```
korikuzma, line 36: This is unrelated to this PR, but I'm kind of wondering if we should change this to `ga4gh_vrs_python` so it's clear that this is not the VRS version?

korikuzma (approved):

Issue #189, opened 2025-07-09
refactor: consolidate registration middleware


jennifer-bowser:
### Feature description

We currently have (or are creating) several middleware functions that only target the `/variation` and `/vrs_variation` registration endpoints. We parse and rebuild the `response` object in each of these middleware functions, and there is a fair amount of duplicate code between them. Based on [this discussion](https://github.com/biocommons/anyvar/discussions/186), it seems it would be a good idea to refactor these functions as follows:

1. Abstract out the main logic of each of these middleware functions into their own utility functions.
2. Create a single middleware function that targets the registration endpoints and runs all the necessary utility functions.

### Use case

- As a developer, it is inconvenient to have to remember to put in checks for each middleware function to ensure the logic only runs for the desired endpoints. Refactoring will streamline the development process for future middleware logic targeting the registration endpoints.
- As a user, I need my applications to be performant. If we continue our current trend of writing new middleware functions every time we want to perform an operation on one of the registration endpoints, we will need to keep parsing/rebuilding the response object and that will eventually start to hurt AnyVar's performance.

### Acceptance Criteria

1. AnyVar has a single middleware function that targets both of the registration endpoints (`/variation` and `/vrs_variation`)
2. The operations currently performed by our various middleware functions that target the registration endpoints are abstracted out into their own functions, which are then called from our new single registration middleware function
3. All relevant test cases are updated as need be to be compatible with the refactor

### Implementation details

The middleware functions that will need to be included in this refactor are:
- `add_creation_timestamp_annotation`
- `add_genomic_liftover_annotation`
- `store_input_payload_annotation`

Don't forget to update their associated tests where applicable.

### Contribution

None

jennifer-bowser:
Note to self: Investigate other types of Response objects that can be parsed without needing to reconstruct it

PR #199, opened 2025-07-24
refactor: consolidate registration middleware (issue #189)

jennifer-bowser:
closes issue #189 

Refactors the middleware logic that adds annotations to newly-registered variants to combine it all into a single middleware function.

I also:
* Extracted out the annotation logic for liftover into its own function, since it's more complicated
* Updated the `get_liftover_variant` function to take a `VrsVariant` class object as a param, instead of a dictionary representation of a variant. This streamlines the logic a little, and it makes more sense (in my mind, at least) in the context of the refactor.


src/anyvar/restapi/main.py
```diff
@@ -310,113 +309,60 @@ def get_variation_annotation(
 
 
 @app.middleware("http")
-async def add_creation_timestamp_annotation(
+async def add_registration_annotations(
     request: Request, call_next: Callable
 ) -> Response:
-    """Add a creation timestamp annotation to a variation if it doesn't already exist."""
+    """Add all required annotations for newly-registered variants"""
```
korikuzma, line 315: It might be good to put in the docstrings what all the required annotations are

src/anyvar/utils/funcs.py
```diff
@@ -19,3 +21,25 @@ def get_nested_key(dict_object: dict, *keys: Any) -> Any:  # noqa: ANN401
...
+    for attr in attributes:
+        try:
+            class_object = getattr(class_object, attr)
+        except AttributeError:
+            return None
+    return class_object
+
+
+def build_vrs_variant_from_dict(variant_dict: dict) -> VrsVariation:
+    """Construct a `VrsVariation` class instance from a dictionary representation of one"""
```
korikuzma, line 43: Missing param + return in docstring

src/anyvar/utils/funcs.py
```diff
@@ -19,3 +21,25 @@ def get_nested_key(dict_object: dict, *keys: Any) -> Any:  # noqa: ANN401
...
     return None
+
+
+def get_nested_attribute(class_object: object, *attributes: Any) -> Any:  # noqa: ANN401
+    """Traverses the path of nested `attributes` in `class_object` and returns the final value.
+    If any key does not exist, returns None.
+
+    :param class_object: The class_object to traverse.
+    :param attributes: The attributes that will be used to traverse the attributes of the class_object, in the order they should be used.
+    :returns: The final value in the property traversal, if all attributes exist; else None.
```
korikuzma, line 32: I think we use return? Idk if it's a big deal, just want to be consistent (looks like there are other places in the codebase that use returns instead of return) 

```suggestion
    :return: The final value in the property traversal, if all attributes exist; else None.
```

src/anyvar/restapi/main.py
korikuzma: Missing param + return in docstring

src/anyvar/utils/liftover_utils.py
```diff
@@ -99,44 +100,44 @@ def _convert_coordinate(converter: Converter, chromosome: str, coordinate: int)
...
-def get_liftover_variant(variant_object: dict, anyvar: AnyVar) -> VrsVariation:
-    """Liftover a variant from GRCh37 or GRCH38 into the opposite assembly, and return the converted variant as a VrsObject.
+def get_liftover_variant(input_variant: VrsVariation, anyvar: AnyVar) -> VrsVariation:
+    """Liftover a variant from GRCh37 or GRCH38 into the opposite assembly, and return the converted variant as a VrsVariation.
     If liftover is unsuccessful, raise an Exception.
 
-    :param variant_object: A dictionary representation of a `VrsVariation`.
-    :param anyvar: An `AnyVar` instance.
-    :return: The converted variant as a `VrsVariation` instance.
+    :param input_variant: A dictionary representation of a `VrsVariation`.
```
korikuzma, line 136:
```suggestion
    :param input_variant: A `VrsVariation`
```

src/anyvar/utils/liftover_utils.py
```diff
@@ -99,44 +100,44 @@ def _convert_coordinate(converter: Converter, chromosome: str, coordinate: int)
...
+    :param anyvar: A `AnyVar` instance.
```
korikuzma, line 137:
```suggestion
    :param anyvar: An `AnyVar` instance.
```

src/anyvar/utils/liftover_utils.py
```diff
@@ -150,14 +151,14 @@ def get_liftover_variant(variant_object: dict, anyvar: AnyVar) -> VrsVariation:
...
-        variant_object, "location", "sequenceReference", "refgetAccession"
+    refget_accession = get_nested_attribute(
+        input_variant, "location", "sequenceReference", "refgetAccession"
     )
-    start_position = get_nested_key(variant_object, "location", "start")
-    end_position = get_nested_key(variant_object, "location", "end")
+    start_position = get_nested_attribute(input_variant, "location", "start")
+    end_position = get_nested_attribute(input_variant, "location", "end")
     if not refget_accession or not start_position or not end_position:
         raise UnsupportedVariantLocationTypeError
```
korikuzma, line 163: Kind of wondering if we could just do:

```suggestion
    try:
        refget_accession = input_variant.location.sequenceReference.refgetAccession
        start_position = input_variant.location.start
        end_position = input_variant.location.end
    except AttributeError:
        raise UnsupportedVariantLocationTypeError
```

and then remove `get_nested_attribute`

src/anyvar/utils/liftover_utils.py
```diff
@@ -203,42 +204,109 @@ def get_liftover_variant(variant_object: dict, anyvar: AnyVar) -> VrsVariation:
...
+            anyvar=anyvar,
+        )
+        # If liftover was successful, we'll annotate with the ID of the lifted-over variant
+        annotation_value = lifted_over_variant.id
+    except LiftoverError as e:
+        # If liftover was unsuccessful, we'll annotate with an error message
+        annotation_value = e.get_error_message()
+
+    # Add the annotation to the original variant
+    annotation_type = "liftover"
```
korikuzma, line 293: This could be a separate issue (if it doesn't already exist), but we might want to consider creating an enum for annotation types we support

korikuzma (changes requested):
Overall looks great! Have some minor suggestions.

src/anyvar/utils/liftover_utils.py
```diff
@@ -203,42 +204,109 @@ def get_liftover_variant(variant_object: dict, anyvar: AnyVar) -> VrsVariation:
...
+            circular=None,
+            sequence=None,
+            moleculeType=None,
+        ),
+    )
 
     # Build the liftover variant object
     # Start by copying the original variant
-    converted_variant_dict = copy.deepcopy(variant_object)
+    converted_variant = copy.deepcopy(input_variant)
```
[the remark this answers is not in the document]
jennifer-bowser: 🤯

src/anyvar/utils/liftover_utils.py
```diff
@@ -203,42 +204,109 @@ def get_liftover_variant(variant_object: dict, anyvar: AnyVar) -> VrsVariation:
...
+    :param original_vrs_id: The ID of the VRS variant to lift over
+    :param original_vrs_object: A dictionary representation of the VRS variant to lift over
+    :param anyvar: An `AnyVar` instance
+    :param annotator: An `AnyAnnotation` instance
+    """
+    # convert `input_vrs_object_dict` into an actual VrsVariation class instance
+    variant_type = input_vrs_variant_dict.get("type", "")
+    input_vrs_variant: VrsVariation = variation_class_map[variant_type](
+        **input_vrs_variant_dict
+    )
```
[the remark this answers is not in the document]
jennifer-bowser: Oh whoops, forgot to update that one! Good catch

src/anyvar/utils/liftover_utils.py
```diff
@@ -203,42 +204,109 @@ def get_liftover_variant(variant_object: dict, anyvar: AnyVar) -> VrsVariation:
...
+            id=None,
+            description=None,
+            aliases=None,
+            extensions=None,
+            residueAlphabet=None,
+            circular=None,
+            sequence=None,
+            moleculeType=None,
+        ),
+    )
```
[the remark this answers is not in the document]
jennifer-bowser: Ohhh I thought Pylance was getting mad at me for not having all the `None` fields. I did think it seemed kinda excessive though! This way is so much nicer 🤣

src/anyvar/utils/liftover_utils.py
```diff
@@ -203,42 +204,109 @@ def get_liftover_variant(variant_object: dict, anyvar: AnyVar) -> VrsVariation:
...
         return_id_obj_tuple=False,
     )
 
-    # return the dereffed lifted-over variant as a VrsObject
+    # return the dereffed lifted-over variant
     dereffed_variant = vrs_deref(o=enreffed_variant, object_store=object_store)
-    return variation_class_map[variant_type](**dereffed_variant.model_dump())
+    return build_vrs_variant_from_dict(
+        dereffed_variant.model_dump()
+    )  # explicitly cast to a VrsVariant so Pylance doesn't get mad
```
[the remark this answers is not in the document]
jennifer-bowser: `vrs_deref` returns a `BaseModel` object, but since I used `VrsVariation` for my return type for this function, Pylance gets mad unless you explicitly cast the `BaseModel` to a `VrsVariation`

src/anyvar/utils/funcs.py
jennifer-bowser: Arg, I forgot these twice in one PR! Nice catch, thank you!

src/anyvar/utils/liftover_utils.py
jennifer-bowser: 🤦‍♀️

src/anyvar/utils/liftover_utils.py
jennifer-bowser: I think the original goal was to purposefully leave annotation types pretty ambiguous so that users can utilize annotations for whatever purposes they need. I think I would be in favor of using an Enum to strictly define annotation types if we still intended to use annotations for information that AnyVar requires for its core functionality (like liftover); but now that we're moving away from that with our storage refactor, I'd vote to keep this flexible.

src/anyvar/utils/liftover_utils.py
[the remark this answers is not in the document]
korikuzma: You can ignore these warnings IMO. 

see https://github.com/ga4gh/vrs-python/blob/586e0c672dd283a99efd57e30326ed6e8c4f14b4/src/ga4gh/vrs/models.py#L531-L560 for defaults being set

src/anyvar/restapi/main.py
```diff
@@ -310,113 +309,65 @@ def get_variation_annotation(
...
-
-        return new_response
+        liftover_annotations = annotator.get_annotation(input_vrs_id, "liftover")
+        if not liftover_annotations:
+            liftover_utils.add_liftover_annotations(
+                input_vrs_id=input_vrs_id,
+                input_vrs_variant_dict=input_variant,
+                anyvar=request.app.state.anyvar,
+                annotator=annotator,
+            )
```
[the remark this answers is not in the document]
jennifer-bowser: Ooh that's an excellent point. I'd say lets punt this to the refactor since at that point we won't be using Annotations for liftover anymore. Also, I believe we intentionally made it so that users could add multiple annotations of the same type if they want, but I 100% agree that our `INSERT` statements should prevent multiple entries for liftover to avoid needing to do these kinds of checks every time.

korikuzma (approved):
Great work 🚀

korikuzma (approved):

jennifer-bowser:
> I concur

@jsstevenson are you by chance familiar with _Catch Me If You Can_?

image

PR #226, opened 2025-08-20
style: update ruff configs

jsstevenson:
Add biocommons template updates -- see https://github.com/biocommons/biocommons.example/pull/57


korikuzma (approved):

korikuzma (approved):
(This is showing up in my feed even though I approved. I think it's after the biocommons infra update)

PR #240, opened 2025-09-15
Moved postgres setup to external sqlalchemy file 

theferrit32:


theferrit32:
oops targeting wrong branch

PR #244, opened 2025-09-16
Add note about XDG_DATA_HOME

theferrit32:
May be overkill to add this but I did hit this and it took me a couple minutes to find out where it was coming from


jsstevenson:
@theferrit32 I have a draft remedy for the above + a few related concerns https://github.com/GenomicMedLab/wags-tails/pull/79

PR #262, opened 2025-10-03
feat: add `mappings` table

jsstevenson:
* add `VariationMappings` table to postgresql implementation

## TODO
* update `nodb`

## Questions
* Provide some sort of description/meta column in mappings table?
* How to handle failed liftover? Previously we would add an annotation with a note about the failed liftover. Is this still something to support?

## Notes
* The existing e2e tests relied upon a mock of the Annotation table. We've talked about trying to move away from that means of testing for maintainability reasons, but there's no immediate replacement because we haven't added mappings CRUD methods to `anyvar.AnyVar` or endpoints to the REST API yet (i.e. there's no way to verify behavior outside of the storage class itself).

### DIFF
--- a/src/anyvar/storage/orm.py
+++ b/src/anyvar/storage/orm.py
@@ -1,6 +1,9 @@
 """SQLAlchemy ORM models for AnyVar database schema."""
 
-from sqlalchemy import ForeignKey, Index, String, create_engine
+from uuid import UUID
+
+from sqlalchemy import ForeignKey, Index, String, create_engine, func
+from sqlalchemy.dialects.postgresql import ENUM as PgEnum  # noqa: N811
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import (
     DeclarativeBase,
@@ -9,6 +12,8 @@ from sqlalchemy.orm import (
     relationship,
     sessionmaker,
 )
+
+from anyvar.storage.base_storage import VariationMappingType
 
 
 class Base(DeclarativeBase):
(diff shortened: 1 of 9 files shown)


src/anyvar/utils/liftover_utils.py
```diff
@@ -232,84 +224,71 @@ def get_liftover_variant(input_variant: VrsVariation, anyvar: AnyVar) -> VrsVari
...
-def add_liftover_annotations(
-    input_vrs_id: str,
-    input_vrs_variant_dict: dict,
-    anyvar: AnyVar,
-) -> None:
+def add_liftover_mapping(variation: VrsVariation, anyvar: AnyVar) -> list[str] | None:
     """Perform liftover between GRCh37 <-> GRCh38. Store the ID of converted variant as an annotation of the original,
-    register the lifted-over variant, and store the ID of the original variant as an annotation of the lifted-over one.
+    register the lifted-over and normalized variant, and store the ID of the original variant as an annotation of
+    the lifted-over one.
```
jennifer-bowser: Since we're not using annotations for this anymore, I think we should update this docstring to remove the references to annotations!

src/anyvar/storage/orm.py
```diff
@@ -91,6 +101,36 @@ class Annotation(Base):
...
+
+    __tablename__ = "variation_mappings"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    source_id: Mapped[str] = mapped_column(
+        String, ForeignKey("alleles.id", ondelete="cascade")
+    )
+    dest_id: Mapped[str] = mapped_column(
+        String, ForeignKey("alleles.id", ondelete="cascade")
+    )
```
jennifer-bowser: I'm not sure Foreign Keys will work here once we have other variant types besides just `Allele`.

jennifer-bowser (changes requested):
When I run all tests, I'm getting errors for `test_storage_implementation.py::test_delete_objects`: 

```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.ForeignKeyViolation) update or delete on table "alleles" violates foreign key constraint "variation_mappings_source_id_fkey" on table "variation_mappings"
```

(Though If we decide to [not use foreign keys for mappings](https://github.com/biocommons/anyvar/pull/262#discussion_r2432693668), this will solve itself)

src/anyvar/utils/liftover_utils.py
jsstevenson: Ugh I goofed the merge conflict resolve here

jennifer-bowser: 😭

Issue #208, opened 2025-07-25
Update the docs


jennifer-bowser:
This should be one of the last tickets in the epic.

Coder should audit all docs pages. Some things to look for include deleting outdated storage options and removing references to optional annotation storage. 

⚠️ Remember to target the `storage-refactor-epic` branch for this

jsstevenson:
> Not sure if we plan to cut all (or some of the readme), but should probably update references to files.

Yeah ideally the readme should be approximately a browser window long imo -- header material, useful links, a quick example, etc. As much as possible, we should only describe how to do something once.

PR #272, opened 2025-10-16
use batch db writes for vcf variant registration (issue #253)

### LINKED ISSUE #253: Update VCF annotator and VCF registrar to batch DB writes with new `Storage` methods
One particular issue is here but the registrar will need other changes to support batched writes

https://github.com/biocommons/anyvar/pull/247#issuecomment-3330204702

jennifer-bowser:
This PR does the following:

1. Renames the `anyvar.put_object` method to `anyvar.put_objects` (plural) and updates it to accept an array of objects instead of just a single object (in line with our implementation of our underlying storage method, `add_objects`).
2. Makes the VCF registration endpoints (`PUT /vcf` and `PUT /annotated_vcf`) use batch database writes. I used a batch size of `10,000`, but we can change this if we want.

I tested this using a ~800MB VCF file with 5.7 million variants. I ran it through the `PUT /vcf` endpoint on both `main` and this feature branch, and the runtimes were _slightly_ slower on this branch but overall seem fairly comparable: 

| | `main` | This branch | Difference vs `main` |
|-|--------|-------------|----------------------------------|
| Runtime (minutes) | 317.12 minutes| 329.95 minutes | +12.83 minutes |
| Runtime (hours) | 5.29 hours | 5.50 hours | +0.21 hours |

I can experiment more with the batch size if we think it could have a meaningful impact on runtimes! Reviewers, I'd love your thoughts on this.

closes #253

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     "ga4gh.vrs[extras]~=2.1.2",
     "sqlalchemy~=2.0.0",
     "pyyaml",
-    "agct",
+    "agct >= 0.2.0rc1",
     "pysam",
     "python-dotenv",
     "anyio",
--- a/src/anyvar/anyvar.py
+++ b/src/anyvar/anyvar.py
@@ -9,8 +9,6 @@ import os
 import warnings
 from urllib.parse import urlparse
 
-from agct import Converter, Genome
-
 from anyvar.storage import DEFAULT_STORAGE_URI
 from anyvar.storage.base_storage import Storage, StoredObjectType
 from anyvar.translate.translate import _Translator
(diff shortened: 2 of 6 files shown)


src/anyvar/queueing/celery_worker.py
```diff
@@ -34,6 +34,11 @@
     broker_url=os.environ.get("CELERY_BROKER_URL", None),
     timezone=os.environ.get("CELERY_TIMEZONE", "UTC"),
     result_expires=int(os.environ.get("CELERY_RESULT_EXPIRES", "7200")),
+    broker_transport_options={
+        int(
+            os.environ.get("CELERY_BROKER_TRANSPORT_OPTIONS_VISIBILITY_TIMEOUT", "7200")
```
jennifer-bowser: This is a new environment variable that we should add to our documentation in #208.

PR #279, opened 2025-10-23
chore: add VCF benchmarking script

### LINKED ISSUE #275: Create benchmark to evaluate prior code to post-storage-refactor code
I would probably use gnomad or 1000 genomes VCF files as the input. Large enough size to get a good idea of real performance and degradation as tables/indexes become non trivial in size.

Use a cloud VM to ensure a consistent environment (e.g. GCP c4a-standard-4). Use a consistent storage device type and size. It may be necessary to run a small test to even estimate what size storage device will be needed for a larger test.

Run postgres natively.

Run AnyVar REST server on localhost.

Upload some amount of VCF records reaching at least tens of millions of variants. It would be interesting to see what happens if loading all 750M variants from gnomAD.

Time the execution time, and record the final postgres data directory size.

Put all this in a bash script / ansible playbook.

Run it on [0.9.1](https://github.com/biocommons/anyvar/releases/tag/0.9.1) and on the latest `storage-refactor-epic`.

jsstevenson:
close #275 

Well, I'm not sure if this closes 275, it depends on what we were thinking for that issue. But this is a thin rewrite of a script I wrote up for GREGoR that can be used to benchmark VCF submissions. It can take multiple input files as arguments and progressively submit them/monitor their progress against a running AnyVar API. There's a multiprocessing part that's there not for benchmarking per se but as a convenience for adding a bunch of big VCFs (eg GREGoR).

Initial unscientific testing on my laptop indicated that the current `storage-refactor-epic` commit performed about as well as `main` on a small (10k records) VCF

## Questions
* Currently polls the status endpoint every 5 seconds. That might be too long

### DIFF
--- a/scripts/vcf_benchmark.py
+++ b/scripts/vcf_benchmark.py
@@ -10,6 +10,8 @@
 import logging
 import time
 from concurrent.futures import ProcessPoolExecutor
+from enum import StrEnum
+from functools import partial
 from json import JSONDecodeError
 from pathlib import Path
 from timeit import default_timer as timer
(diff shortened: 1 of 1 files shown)


scripts/vcf_benchmark.py
```diff
@@ -0,0 +1,108 @@
...
+
+def submit_variants(file: Path, anyvar_host: str) -> None:
+    with file.open("rb") as f:
+        response = requests.put(
+            f"{anyvar_host}/vcf",
+            files={"vcf": (file.name, f, "text/plain")},
+            params={
+                "allow_async_write": True,
+                "run_async": True,
+                "require_validation": False,
```
jennifer-bowser: It would be nice if we could set these params + the other params this endpoint supports (e.g., `for_ref`, `assembly`, etc.) ourselves when calling this script rather than hard-coding them here.

Not necessarily a requirement for right now though if we'd rather implement that later.

jennifer-bowser (changes requested):

jennifer-bowser:
Interesting - I tried using this script on a VCF file with about 5.7 million variants, and the script indicated that the run was complete several hours before the run actually finished? I'm not really sure why. It worked really well for me on a smaller file though (400,000 variants).

scripts/vcf_benchmark.py
```diff
@@ -0,0 +1,157 @@
...
+                    filename = f"vcf_benchmark_output_{run_id}"
+                    with Path(filename).open("w") as f:
+                        f.write(response.text)
+                    logger.info(
+                        "Received malformed response from vcf run ID lookup. Saved to %s",
+                        filename,
+                    )
+                return
+    # don't need to poll for sync response
+    # timeouts should be expected for larger files though
```
jennifer-bowser: I did a synchronous run of a 400,000 variant file (~30MB) and it timed out pretty quickly. Is there a way to up the timeout cutoff?

jsstevenson:
> the script indicated that the run was complete several hours before the run actually finished

Well, I give up

> I did a synchronous run of a 400,000 variant file (~30MB) and it timed out pretty quickly. Is there a way to up the timeout cutoff?

Yeah, it's just the variable set on line 54. In general, the synchronous endpoint is really not intended for truly large files, which is why the cutoff is relatively low -- you'd want to know quickly if something broke. But if something broke, you'd probably just get an immediate 500. Maybe this could be a command line configuration and/or maybe the default should be much higher?

jennifer-bowser:
> the script indicated that the run was complete several hours before the run actually finished

Oh you know what, this isn't an issue with your script - it's an AnyVar issue. I tried running this again and this time I realized that the poll endpoint itself is returning a status of `NOT FOUND` even though the run is still actually going. 

So that's definitely a problem, just not one for this PR 🤣

jennifer-bowser (approved):

PR #292, opened 2025-11-03
Re-add `postgres_init.sql` files

jennifer-bowser:


jsstevenson:
A few thoughts

1) I would personally vote against checking files meant for testing into `src/`
2) personally I think it's easier to set this stuff via env vars on the container or (since it's short) just a single `psql` call
3) we'll need to make sure that package data is declared in pyproject.toml if it should be included in the distributed source https://stackoverflow.com/questions/69647590/specifying-package-data-in-pyproject-toml

korikuzma (commented):
@jsstevenson already reviewed this

Issue #304, opened 2025-11-07
Getting an annotation via `GET /variation/{vrs_id}/annotations/{variation_type}` for a VRS ID that doesn't exist returns `200 OK`


jsstevenson:
```pycon
>>> import requests
>>> response = requests.get("http://localhost:8000/variation/vrs_id_that_doesnt_exist/annotations/doesnt_matter")
>>> response.status_code
200
>>> response.json()
{'annotations': []}
```

I don't think this is intentional? It's inconsistent with behavior in `POST /variation/{vrs_id}/annotations`

PR #309, opened 2025-11-07
test: add annotation endpoint tests

### LINKED ISSUE #305: Add tests for annotation endpoints

jsstevenson:
close #305 

There are a few things, like behavior in #304 as well as stuff like updating/deleting that we have planned but are not immediate priorities, that I didn't cover here, so this ends up being pretty basic


korikuzma (approved):
Minor, but consider adding case for invalid pyaload

PR #315, opened 2025-11-13
refactor: streamline supported vrs object types (issue #265)

### LINKED ISSUE #265: Remove 'StoredObjectType' enum in favor of 'VrsObject' enum
When I originally created our refactor class diagram, I envisioned us using the `object` database CRUD methods for nearly everything - Annotations, Alleles and other VRS Variants, SequenceReferences, Mappings, etc. We created a `StoredObjectType` enum to classify what kind of object we were storing with this method.

However, we've moved away from this paradigm in favor of using the `object` methods only for VRS objects, with other types of items (e.g., Annotations and Mappings) having their own dedicated CRUD methods. After removing those from the `StoredObjectType` enum, it's now pretty much equivalent to our pre-existing `VrsObject` enum. There's no need for two enums that do the same thing, so we should nix the `StoredObjectType` enum and just use the `VrsObject` one.

This will involve:
- [ ] Deleting the `StoredObjectType` enum definition from `base_storage.py`
- [ ] Adding `SequenceReference` to the `VrsObject` enum defined in `types.py`
- [ ] Updating all references to `StoredObjectType` (including method signatures) to use `VrsObject` instead.

jennifer-bowser:
closes #265 

### Overview

I streamlined our `VrsObject` and `VrsVariation` types and related enums so that we have one source of truth for each. 

The biggest updates in this PR are in `src/anyvar/utils/types.py`: 
* I got rid of the `StoredObjectType` enum so that we're just using the `VrsObject` type union instead.
* I wanted to get rid of our `SupportedVariationType` enum, but we really do need this for our `RegisterVariationRequest` model; so instead I am creating the enum dynamically based on the models listed in our `VrsVariation` type union
* I also updated our `vrs_object_class_map` to be generated dynamically so we don't need to manually add things to it each time we want to support a new type of Vrs Object.

### TLDR:
Previously, if we were to add support for a new type of VRS Variation, we'd have to add it in 5 places:
* The `StoredObjectType` enum
* The `VrsObject` type union
* The `VrsVariation` type union
* The `SupportedVariationType` enum
* The `variation_class_map` dict

Now, we'll only have to update 2 places:
* The `VrsObject` type union
* The `VrsVariation` type union

### DIFF
--- a/src/anyvar/utils/types.py
+++ b/src/anyvar/utils/types.py
@@ -22,6 +22,7 @@ VrsVariation = models.Allele | models.CopyNumberChange | models.CopyNumberCount
 
 class SupportedVariationType(StrEnum):
     """Supported variation types for API input. Enum is dynamically built from the models in the `VrsVariation` type union.
+    This should only be used to parse HTTP input in the `src/anyvar/restapi/schema.py` models, not within application logic.
 
     Example:
     >>> SupportedVariationType.COPY_NUMBER_CHANGE = "CopyNumberChange"


src/anyvar/utils/types.py
```diff
@@ -19,20 +17,33 @@
...
+
+
+"""
+Builds a dict in the form of `"ModelName": models.ModelName` for every class listed in the `VrsObject` type union
+For example:
+>>> vrs_object_class_map["Allele"] = models.Allele
+"""
+vrs_object_class_map: dict[str, type[VrsObject]] = {
+    cls.__name__: cls for cls in get_args(VrsObject)
+}
```
jennifer-bowser: The only place this is used right now is in `vrs_object_utils.py`. I'm not sure if it should stay here, or move to live in that file. I'd love reviewer's thoughts!

src/anyvar/utils/vrs_object_utils.py
```diff
@@ -0,0 +1,13 @@
+"""Util functions related to vrs objects"""
```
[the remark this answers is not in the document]
jennifer-bowser: Ok @jsstevenson I've just moved these to a file in the `tests` directory called `test_utils.py`, take a look and see what you think!

tests/test_utils.py
```diff
@@ -0,0 +1,20 @@
+from ga4gh.vrs import models
```
jsstevenson: okay, very sorry, this is very nitpicky, but I would expect a file named "tests/test_utils.py" to be for tests performed on utils -- could you rename, or move to `conftest.py`?

jsstevenson (changes requested):

tests/test_utils.py
jennifer-bowser: I never mind nitpicky things! They make our codebase better 😄

PR #318, opened 2025-11-14
refactor: simplify add objects method (Issue #268)

### LINKED ISSUE #268: Simplify `add_objects` method and push object model disassembly either into sqlalchemy or our ORM classes
Right now the add_objects method in postgres.py is big because it manually pulls apart the different objects and calls `.execute` to insert a list of dicts of each object type. The sqlalchemy `merge` function can do this for us but I was finding it to be significantly slower. We should look more into that because using `merge` would be ideal.

If not, we should remove the object-type-specific code from `add_objects` and push it into orm.py, with a method on each ORM object type.

We used a similar pattern in `clinvar-ingest` where each object knows how to disassemble itself into its constituent parts.

Example:
https://github.com/clingen-data-model/clinvar-ingest/blob/9d7ea2d7db9e200556c773d7f04877dca1968dbc/clinvar_ingest/model/variation_archive.py#L761-L770

jennifer-bowser:
closes #268

### DIFF
--- a/src/anyvar/storage/postgres.py
+++ b/src/anyvar/storage/postgres.py
@@ -108,7 +108,7 @@ class PostgresObjectStore(Storage):
 
             object_parts = db_entity.disassemble()
             for entity_type, entity in object_parts.items():
-                vrs_objects[entity_type][entity.id] = entity
+                vrs_objects[entity_type][entity.id] = entity  # type: ignore (all children of orm.Base have an `id`)
 
         with self.session_factory() as session, session.begin():
             # Use ON CONFLICT DO NOTHING to handle duplicates gracefully
--- a/src/anyvar/storage/orm.py
+++ b/src/anyvar/storage/orm.py
@@ -47,7 +47,7 @@ class Base(DeclarativeBase):
             column.name: getattr(self, column.name) for column in self.__table__.columns
         }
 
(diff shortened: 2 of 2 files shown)


src/anyvar/storage/postgres.py
```diff
@@ -85,67 +86,44 @@ def add_objects(self, objects: Iterable[types.VrsObject]) -> None:
...
+
+        # Sort objects so that we insert in dependency order: sequence_references -> locations -> alleles
+        insert_order: list[str] = [
+            orm.SequenceReference.__name__,
+            orm.Location.__name__,
+            orm.Allele.__name__,
+        ]
+        sorted_vrs_objects: dict[str, dict[str, orm.Base]] = {
+            key: vrs_objects[key] for key in insert_order if key in vrs_objects
+        }
```
jsstevenson: Any reason why we can't do this the previous way (with separate variables) or in a dict with predefined fields?

jsstevenson: Oh, also, it seems like in some cases sqlalchemy can handle the dependency order itself within a session if we are declaring relationships properly? do we know that we need to be doing this sorting stuff on our side?

jsstevenson (changes requested):

jsstevenson (approved):
awesome!

PR #322, opened 2025-11-20
fix: return 404 when resource not found

### LINKED ISSUE #304: Getting an annotation via `GET /variation/{vrs_id}/annotations/{variation_type}` for a VRS ID that doesn't exist returns `200 OK`

jsstevenson:
close #304 

Return a 404 if the VRS ID in `/variation/<vrs_id>/annotations` or `/variations/<vrs_id>/mappings` doesn't exist.

If the `ObjectStorage.get_annotations()` lookup returns an empty list, *then* do a check to see if the VRS ID exists at all. If it doesn't, then return the error/a 404. 

### Questions/thoughts

* It feels weird to do this for mappings ONLY if the source ID doesn't exist (not the destination ID), but IMO that seems like the most RESTy way. If instead the path was something like `GET /variation/<source_id>/mappings/<dest_id>/` then I think I'd be more comfortable trying to do this for both.

### DIFF
--- a/src/anyvar/storage/mapper_registry.py
+++ b/src/anyvar/storage/mapper_registry.py
@@ -51,7 +51,7 @@ class MapperRegistry:
         mapper = self.get_mapper(type(db_entity))
         return mapper.from_db_entity(db_entity)
 
-    def to_db_entity(self, anyvar_entity):  # noqa: ANN201, ANN001
+    def to_db_entity(self, anyvar_entity) -> orm.Base:  # noqa: ANN001
         """Convert any VRS model to its corresponding DB entity."""
         # Map VRS model types to DB entity types
 
--- a/src/anyvar/storage/orm.py
+++ b/src/anyvar/storage/orm.py
@@ -2,6 +2,7 @@
 
 import os
 import re
+from collections.abc import Iterator
 
 from sqlalchemy import (
     ForeignKey,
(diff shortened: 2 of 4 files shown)


src/anyvar/anyvar.py
```diff
@@ -208,13 +219,20 @@ def get_object_mappings(
...
         except Exception:
             _logger.exception(
                 "Failed to retrieve mappings for source_object_id: %s and mapping_type: %s",
                 source_object_id,
                 mapping_type,
             )
             raise
+        if not mappings:
+            try:
+                _ = self.get_object(source_object_id)
```
jennifer-bowser: super-duper minor nitpick, but in the `get_object_annotations` above we're calling `self.get_object(object_id)` without assigning it to any variable, and here we're doing the same thing but assigning the result to `_`. I think either is totally fine, but it'd be nice to be consistent and do it the same way in both places.

jennifer-bowser (approved):
I have one suggestion but it's super minor and optional, so I'm approving!

PR #327, opened 2025-11-24
feat: use enum for molecule_type in sequence_references table

### LINKED ISSUE #311: `molecule_type` column in `sequence_references` table should be an enum
From @jennifer-bowser comment here https://github.com/biocommons/anyvar/pull/271#discussion_r2503681529

This is supposed to correspond to the VRS `MoleculeType`:

```python
class MoleculeType(str, Enum):
    """Molecule types as `defined by RefSeq <https://www.ncbi.nlm.nih.gov/books/NBK21091/>`_ (see Table 1)."""

    GENOMIC = "genomic"
    RNA = "RNA"
    MRNA = "mRNA"
    PROTEIN = "protein"
```

but currently it's just a string

```python
    molecule_type: Mapped[str | None]
```

korikuzma:
close #311

```shell
anyvar=# \dT+ moleculetype;
                                         List of data types
 Schema |     Name     | Internal name | Size | Elements | Owner  | Access privileges | Description 
--------+--------------+---------------+------+----------+--------+-------------------+-------------
 public | moleculetype | moleculetype  | 4    | genomic +| anyvar |                   | 
        |              |               |      | RNA     +|        |                   | 
        |              |               |      | mRNA    +|        |                   | 
        |              |               |      | protein  |        |                   | 
(1 row)

anyvar=# INSERT INTO sequence_references (id, molecule_type) VALUES ('id', 'protein');
INSERT 0 1
anyvar=# INSERT INTO sequence_references (id, molecule_type) VALUES ('id', 'PROTEIN');
ERROR:  invalid input value for enum moleculetype: "PROTEIN"
LINE 2: VALUES ('id', 'PROTEIN');
```

### DIFF
--- a/src/anyvar/storage/orm.py
+++ b/src/anyvar/storage/orm.py
@@ -13,7 +13,6 @@ from sqlalchemy import (
     UniqueConstraint,
     create_engine,
 )
-from sqlalchemy.dialects.postgresql import ENUM as PgEnum  # noqa: N811
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import (
     DeclarativeBase,
@@ -143,8 +142,10 @@ class SequenceReference(Base):
     molecule_type: Mapped[MoleculeType | None] = mapped_column(
         Enum(
             MoleculeType,
+            name="molecule_type",
             native_enum=True,
             values_callable=lambda mt_enum: [mt.value for mt in mt_enum],
+            validate_strings=True,
         )
     )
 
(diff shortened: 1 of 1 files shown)


jsstevenson (changes requested):
👍 looks good but could you update the other enum column to use the generic enum type with `use_native` as well?

I think for both we want to provide `name` args ("molecule_type" and "mapping_type" or something) and `validate_strings=True` 

so

```python
        Enum(
            MoleculeType,
            name="molecule_type",
            native_enum=True,
            values_callable=lambda mt_enum: [mt.value for mt in mt_enum],
            validate_strings=True
        )
```

PR #328, opened 2025-11-24
feat!: add BATCH_SIZE (n=100) limit to PostgresObjectStore

### LINKED ISSUE #294: Add `LIMIT` param to all SQL queries
### Feature description

Currently, AnyVar could crash if users attempt to run queries that return a very large number of records (example - hitting the `GET /search` endpoint with a `start` and `end` range that encompasses an entire chromosome). Eventually, we will want to solve this problem by paginating our responses. However, in order to ship our MVP release on time, we've opted to simply limit the number of records an individual SQL query can return for now.

### Use case

I am frustrated when AnyVar crashes because it can't support large queries

### Acceptance Criteria

SQL queries are limited to a reasonable number of records so that AnyVar does not crash when attempting to run very large queries

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Implementation details

_No response_

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

None

korikuzma:
close #294

We can bump the batch size, just randomly chose a number

### DIFF
--- a/src/anyvar/storage/postgres.py
+++ b/src/anyvar/storage/postgres.py
@@ -32,6 +32,8 @@ class PostgresObjectStore(Storage):
     with object mapping to convert between VRS models and database entities.
     """
 
+    BATCH_SIZE = 100
+
     _VRS_OBJECT_INSERT_ORDER: list[str] = [  # noqa: RUF012
         orm.SequenceReference.__name__,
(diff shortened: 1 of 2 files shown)


jsstevenson (approved):
👍 100 is good. This is intended to basically act as a placeholder until we work out pagination

PR #346, opened 2025-12-04
refactor: move timestamp annotation to anyvar class + add unit tests

### LINKED ISSUE #223: refactor: move logic out of API endpoint handlers
### Feature description

We have a lot of logic living directly in our endpoint handlers in `main.py`. @jsstevenson brought up the excellent point that this paradigm is something we should move away from. There are several reasons for this:

1. **It will help future-proof AnyVar:**
    - We currently can't access AnyVar functionality through any means _other_ than a REST API. If/when we want to allow access to AnyVar in other ways (e.g., through a local Python installation, etc.), we would need to do this refactor first
2. **It makes our test logic cleaner:**     
    - Separating out our endpoint handlers and our core logic allows us to test each separately, which is what we should be doing anyways to align with best practices.

### Use case

We want AnyVar to be a maintainable, bug-free system that can easily grow to fit our needs now and in the future. This refactor will help us in that goal.

### Acceptance Criteria

* Endpoint handlers contain minimal logic. All core functionalities should be abstracted out into their own functions
* Tests are updated/created for both the endpoint handlers and the abstracted functionality as needed

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Implementation details

_No response_

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

None

korikuzma:
close #223

### DIFF
--- a/src/anyvar/anyvar.py
+++ b/src/anyvar/anyvar.py
@@ -119,7 +119,7 @@ class AnyVar:
     def get_object(
         self, object_id: str, object_type: type[types.VrsObject] | None = None
     ) -> VrsObject:
-        """Retrieve registered variation.
+        """Retrieve registered VRS Object.
 
         :param object_id: object identifier
         :param object_type: specific object type to search (optional - if not provided, searches all types)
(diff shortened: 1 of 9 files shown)


korikuzma:
@jsstevenson had to resolve merge conflict

jsstevenson (approved):

PR #364, opened 2025-12-29
Change db column name mapping for location to avoid ANSI SQL reserved words

ehclark:
The `Location` entity has a `start` column.  `START` is a reserved word in ANSI SQL and better avoided.

Map `start` and `end` to `start_pos` and `end_pos` columns in the database to avoid reserved words.


jsstevenson (approved):
👍

Issue #342, opened 2025-11-25
Store vrs_objects table in addition to the tables per type


korikuzma:
https://github.com/biocommons/anyvar/blob/52231711068daf192a33705cb606ccd71b87f686/src/anyvar/storage/postgres.py#L78-L95

Originally added by @theferrit32

jennifer-bowser:
We could potentially make this a view rather than an actual table

PR #367, opened 2026-01-02
Re-implement support for Snowflake with new storage API

ehclark:
Created an implementation of the `Storage` API that allows AnyVar to use Snowflake as a backend database.

Some highlights worth a mention here:
- SQLAlchemy ORM mappings and query compilation customizations were needed to work with the Snowflake:
    - Data type override needed for the `dict` columns so they are treated converted to/from `str` on the client side
    - Customization of `INSERT` statements were required to convert strings into Snowflake `VARIANT` data type
    - Snowflake has no `ON CONFLICT DO NOTHING`, so instead there is an option (on by default) to rewrite bulk inserts to include an outer join to the target table to avoid duplicate keys.  For `VariationMapping` single row inserts, a merge statement is used, since volume is expected to be low (at least for now).
    - Snowflake does not support indexes on regular tables.  Added logic to to omit indexes from schema definition when dialect is `snowflake`.
- As a Snowflake specific optimization, there is an option to use dynamic tables.  Instead of running 3 inserts for each `Allele` record, there is a single write to the `VrsObject` table and then dynamic tables (similar to materialized views) are used for the `Allele`, `Location` and `SequenceReference` tables.  So this effectively implements https://github.com/biocommons/anyvar/issues/342 in the Snowflake connector.
- Performance is comparable to the current 0.9.1 version without needing asynchronous database writes, so the Snowflake backend is synchronous for now.
- For unit tests, factored out low level re-usable functions that test the storage API into `tests/unit/storage/storage_test_funcs.py` and then call those functions from `test_postgres.py` and `test_snowflake.py`.  Added a pytest mark to run postgres or snowflake tests only depending on the database being used in the test run.
- Snowflake cannot currently be an optional dependency because of the necessary db type override in `orm.py`.  References to the Snowflake libraries are required as part of the schema metadata even when Snowflake is not being used.  I tried to find a way around this, but could not.

The default behavior is as OLTP-like as is reasonable.  MGB will run this using dynamic tables and no protection against duplicate keys to maximize throughput (the database this is integrated with already assumes duplicate keys will be present and has necessary protections).

Also fixes https://github.com/biocommons/anyvar/issues/368

### DIFF
--- a/src/anyvar/restapi/main.py
+++ b/src/anyvar/restapi/main.py
@@ -41,6 +41,7 @@ from anyvar.restapi.schema import (
     RegisterVariationResponse,
     SearchResponse,
     ServiceInfo,
+    TranslationResult,
     VariationRequest,
 )
 from anyvar.restapi.vcf import router as vcf_router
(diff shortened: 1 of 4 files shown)


src/anyvar/storage/orm.py
```diff
@@ -135,37 +166,52 @@ def get_disassembler(self) -> Iterator[Base]:
...
-            "object_id",
-            "annotation_type",
-        ),
-    )
+    @declared_attr
+    @classmethod
+    def __table_args__(cls):  # noqa: ANN206
+        uri = os.environ.get("ANYVAR_STORAGE_URI", DEFAULT_STORAGE_URI)
+        parsed_uri = urlparse(uri)
+        if parsed_uri.scheme == "snowflake":
```
jsstevenson: Maybe it would be worth refactoring this out so that this value is calculated once between each ORM class? I imagine we'll be creating more indexes on the postgres side in the future

jsstevenson (approved):
👍 looks good, it appears that effects on existing behavior are negligible at best so if this works for you, I'm fine with merging as-is.

* I will say that if the dynamic tables stuff works for you, I think we can just close https://github.com/biocommons/anyvar/issues/342 (I believe we opened this ticket to preserve functionality for the MGB use case)
* I wonder if we should split out postgres and snowflake dependencies into their own groups at some point, not sure if imposes any headaches to install them both by default when you would presumably only ever need one

jsstevenson (approved):

PR #416, opened 2026-03-24
feat: add gene-based search

### LINKED ISSUE #250: Implement search by Gene IDs and Transcript IDs
Reference annotations are in the NCBI GFF files, e.g.
https://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/000/001/405/GCF_000001405.26_GRCh38/

This data may already be made available via UTA or some other tool in our ecosystem. But we could re-load this data, because it is small.

Load MANE annotations too?

jsstevenson:
close #250 

Currently just fetches matches on newest reference build (i.e. likely GRCh38). Future would could include specification of a reference, or also grabbing stuff on older builds too (need to think about pagination funkiness).

### DIFF
--- a/src/anyvar/features/genes.py
+++ b/src/anyvar/features/genes.py
@@ -1,0 +1,51 @@
+"""Provide tools for gene-based searches.
+
+Use in contexts like gene-based searches; functions here provide coordinates, which can
+be fed to core AnyVar tables for variant retrieval.
+"""
+
+from typing import NamedTuple
+
+from asyncpg.pool import PoolConnectionProxy
+
+
+class GeneCoords(NamedTuple):
+    """Data class for gene coords lookup result"""
+
+    hgnc_symbol: str
+    accession: str
+    start_i: int
+    end_i: int
+
+
+async def get_gene_coords(
+    uta_conn: PoolConnectionProxy, gene: str
+) -> GeneCoords | None:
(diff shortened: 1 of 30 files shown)


korikuzma:
I believe @jsstevenson said in standup that this is pending [this Cool-Seq-Tool PR](https://github.com/GenomicMedLab/cool-seq-tool/pull/465)

jsstevenson:
It was in draft in my brain

src/anyvar/features/genes.py
```diff
@@ -0,0 +1,51 @@
...
+from typing import NamedTuple
+
+from asyncpg.pool import PoolConnectionProxy
+
+
+class GeneCoords(NamedTuple):
+    """Data class for gene coords lookup result"""
+
+    hgnc_symbol: str
+    acc: str
```
korikuzma: Consider using the full word or `ac` (which I think is consistent with our other tooling). I think I prefer full word though

```suggestion
    accession: str
```

korikuzma (approved):

src/anyvar/features/genes.py
jsstevenson: yeah agreed

korikuzma (approved):

korikuzma (approved):

korikuzma (approved):

Issue #259, opened 2025-10-02
Fix pyright warnings and add to CI


theferrit32:
A number of type warnings were identified in the review of this PR

https://github.com/biocommons/anyvar/pull/247

We should also add this to the CI github action and maybe to a precommit hook too, if this is something we want to start enforcing.

jsstevenson:
Also: add `typed` classifier https://pypi.org/search/?c=Typing+%3A%3A+Typed

Could enable adding a shields.io badge https://img.shields.io/pypi/types/anyvar

PR #420, opened 2026-04-08
feat: adds bicommons metadata to service-info endpoint

### LINKED ISSUE #417: Add biocommons metadata to `/service-info`
### Feature description

We currently store Spec (VRS) and Reference Implementation (VRS-Python) versions. We should consider adding SeqRepo and UTA versions that are being used. 

### Use case

* I want to know what versions of SeqRepo and UTA are being used. This could impact results that I am expecting

### Acceptance Criteria

Given I want to know what versions of SeqRepo and UTA are being used,
When I make a request to `GET /service-info`,
Then I should receive a response then I should see something like:

```json
"impl_metadata": {
    "vrs_python_version": "2.2.0",
    "seqrepo_version": "2024-12-20",
    "uta_schema": "uta_20241220"
  }
```

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Implementation details

_No response_

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

None

jennifer-bowser:
closes #417

To test, go to `main.py` > `app_lifespan` and comment out ~line 53 that defines `service_info_config_file`; initialize it to `None` instead : 
```python
# service_info_config_file = os.environ.get("ANYVAR_SERVICE_INFO")
service_info_config_file = None
```

### DIFF
--- a/.service_info_example.yaml
+++ b/.service_info_example.yaml
@@ -13,6 +13,5 @@ createdAt: 2025-06-01T00:00:00Z
 updatedAt: 2025-06-01T00:00:00Z
 environment: dev
 impl_metadata:
-  vrs_python_version: "2.3.1"
   seqrepo_version: "2024-12-20"
   uta_schema: "uta_20241220"
--- a/.service_info_example.yaml
+++ b/.service_info_example.yaml
@@ -12,3 +12,7 @@ documentationUrl: https://github.com/biocommons/anyvar
 createdAt: 2025-06-01T00:00:00Z
 updatedAt: 2025-06-01T00:00:00Z
 environment: dev
+impl_metadata:
+  vrs_python_version: "2.3.1"
+  seqrepo_version: "2024-12-20"
+  uta_schema: "uta_20241220"
--- a/src/anyvar/restapi/schema.py
+++ b/src/anyvar/restapi/schema.py
@@ -97,11 +97,12 @@ class ImplMetadata(BaseModel):
         Example:
         >>> os.environ.setdefault(
         ...     "SEQREPO_DATAPROXY_URI",
-        ...     "seqrepo+file:///usr/local/share/seqrepo/2024-12-20"
+        ...     "seqrepo+file:///usr/local/share/seqrepo/2024-12-20",
         ... )
         >>> path_name = ImplMetadata._get_env_var_path_name("SEQREPO_DATAPROXY_URI")
         >>> print(path_name)
         "2024-12-20"
+
         """
         path = Path(os.environ.get(env_var) or "")
         return path.name or None
--- a/src/anyvar/restapi/schema.py
+++ b/src/anyvar/restapi/schema.py
@@ -1,11 +1,12 @@
 """Provide response definitions to REST API endpoint."""
 
+import os
 from collections.abc import Iterable
 from datetime import datetime
 from enum import Enum, StrEnum
 from typing import Any
 
-from biocommons.seqrepo import __version__ as seqrepo_version
+from anyio import Path
 from ga4gh.vrs import (
     VRS_VERSION,
     VrsType,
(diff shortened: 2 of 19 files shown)


src/anyvar/restapi/schema.py
```diff
@@ -5,6 +5,7 @@
 from enum import Enum, StrEnum
 from typing import Any
 
+from biocommons.seqrepo import __version__ as seqrepo_version
```
jennifer-bowser: Question: this seqrepo version is in the format `"0.6.11"`; however, the example in the ticket for this uses a date for the seqrepo version (`"2024-12-20"`). Which format do we want? 

If we want the date version, I'll remove this line and have users set the seqrepo version via the `ANYVAR_SERVICE_INFO` file instead, just like the `uta_schema`.

jsstevenson: I think the intent was to capture the seqrepo snapshot version (I think we should be precise about the property name). I am concerned that there isn't a way to reliably and accurately capture this value, though. Maybe to be pragmatic we could do something like

1) try to grab it from the end of `SEQREPO_DATAPROXY_URI` of the form `\d\d\d\d-\d\d-\d\d`
2) if that fails, check if there's a value provided in the config yaml
3) if not, then leave it as null

jennifer-bowser: I tried to do something similar to option 1 for the `uta_schema`, but I ran into a problem where the `schema.py` file is initialized before the `load_dotenv` call in `main.py`, so reading from environment variables to set class attributes doesn't work here 😭 

We could pivot to creating a config file that creates a Pydantic `Settings` class to house env vars, like we do in AnyVLM; but I wasn't sure we wanted to undertake that big of a refactor for this ticket. I'm down to do it though if we want to go that direction!

jsstevenson: I like pydantic settings and would never say no to adding it to a project

That being said, I think it makes sense that these values are set at app initialization rather than at schema module import (we get VRS python version from the schema class definition itself since that's effectively a const defined by VRS-Python, not really anything you would need to wait to determine until runtime) -- can we just attempt to populate from env vars in `app_lifespan()`?

ideally the seqrepo class would be able to tell us this itself, but clearly, we live in an imperfect world

jennifer-bowser: Ahh, good idea; I like that approach! I'll go ahead and set these during app initialization

jennifer-bowser: Update: I found a way to do what I originally intended, which lets these values be set when the class is instantiated. It's a bit more wordy, but I think it works ok for our purposes. Let me know your thoughts. I'm open to changing this up if we don't like it!

.service_info_example.yaml
```diff
@@ -12,3 +12,7 @@ documentationUrl: https://github.com/biocommons/anyvar
 createdAt: 2025-06-01T00:00:00Z
 updatedAt: 2025-06-01T00:00:00Z
 environment: dev
+impl_metadata:
+  vrs_python_version: "2.3.1"
```
jsstevenson: it occurs to me that this should not be a configurable field, but that's probably a low priority to fix in the future

src/anyvar/restapi/schema.py
```diff
@@ -75,6 +77,36 @@ class ImplMetadata(BaseModel):
...
+        Example:
+        >>> os.environ.setdefault(
+                "SEQREPO_DATAPROXY_URI",
+                "seqrepo+file:///usr/local/share/seqrepo/2024-12-20"
+            )
+            path_name = _get_env_var_path_name("SEQREPO_DATAPROXY_URI")
+            print(path_name)
+            ---
+            "2024-12-20"
+
```
jsstevenson, line 110: this formatting is a bit more likely to render properly, also make sure the function is called correctly

```suggestion
        >>> os.environ.setdefault(
        ...     "SEQREPO_DATAPROXY_URI",
        ...     "seqrepo+file:///usr/local/share/seqrepo/2024-12-20"
        ... )
        >>> path_name = ImplMetadata._get_env_var_path_name("SEQREPO_DATAPROXY_URI")
        >>> print(path_name)
        "2024-12-20"
```

jsstevenson (changes requested):
small tweak

.service_info_example.yaml
jennifer-bowser: Hmm, good point. I think it'd be quick to throw something in here that prevents it from being overridden! Let me see if I can knock that out. If for some reason it's not as easy as I think it should be, we can bump it to a future ticket

jsstevenson (dismissed):

src/anyvar/restapi/schema.py
```diff
@@ -86,6 +86,14 @@ class ImplMetadata(BaseModel):
...
     )
 
+    @model_validator(mode="before")
+    @classmethod
+    def forbid_vrs_python_override(cls, data: dict | None) -> dict | None:
+        """prevent 'vrs_python_version' from being overridden"""
+        if isinstance(data, dict) and "vrs_python_version" in data:
+            raise ValueError("`vrs_python_version` cannot be overridden")
+        return data
+
```
korikuzma: I think we should probably go with a [computed field property](https://pydantic.dev/docs/validation/2.0/usage/computed_fields/) instead

jennifer-bowser:
Ah noooo I just pushed changes for #259 to this branch by accident 😭 Hang on, I'll revert that

src/anyvar/restapi/schema.py
jennifer-bowser: I originally considered a computed field, but went with the model validator so that it would throw an explicit error if someone attempt to override it. I like the transparency of throwing an error (so people don't go "hey, why isn't my override working?") 

That said, a computed field would definitely be more straightforward and feels a little less hacky. I'm open to either! I can update this to be a computed field if we want.

jsstevenson (approved):

korikuzma (approved):

Issue #146, opened 2025-05-02
Write functions for forward and reverse transcription and translation


ahwagner:
### Feature description

Going from g. <-> (preferred) c. <-> p. representations need corresponding methods. These are used in the Variation Normalizer, and need to be extracted for reuse in this product.

### Use case

For us to match across evidence and searched contexts in AnyVar queries, we will need to be able to transform across these three levels of sequence variants from the central dogma.

### Acceptance Criteria

We can forward transform variants from g. to c. to p.
NOTE: return with specific example inputs and outputs

We can reverse transform variants from c. to g.
NOTE: return with specific example inputs and outputs

We can reverse transform coordinates from p. to c. to g. contexts.
NOTE: return with specific example inputs and outputs

### Proposed solution

Use the MANE Select / Plus Clinical / Longest transcript selection criteria used by cool-seq-tool for application to this issue. Track the version of the MANE resource used for this for provenance.

https://coolseqtool.readthedocs.io/0.13.1/transcript_selection.html

### Contribution

None

korikuzma:
https://coolseqtool.readthedocs.io/stable/reference/api/mappers/cool_seq_tool.mappers.mane_transcript.html#cool_seq_tool.mappers.mane_transcript.ManeTranscript.get_mane_transcript

ahwagner:
1. We want to be able to support the generic operations above:
   1. As Python functions
   2. As REST endpoints
2. For g->c (transcript context mappings), c->p (protein context mapping), and c->g operations (genomic context mappings):
   1. If the input variant is already registered, these mappings (and their corresponding output variants) should also already be stored in AnyVar and simply retrieved
   2. If the input variant is not already registered, we should register it, compute these mappings, and store the mappings and output variants.
3. For p->c, I think we might want to dig a little deeper into design choices. Will come back to this later.

korikuzma:
I have some questions... @biocommons/anyvar-maintainers 

* I know a while ago we had the Variation Normalizer as a translator. I can't remember why we removed it. Did we want to add this back or only rely on Cool-Seq-Tool/VRS-Python? At the moment, Cool-Seq-Tool does not return VRS objects, but we can change that if this way is preferred.
* What nomenclature do we need to support (gnomad vcf, hgvs, spdi, and/or others)?

theferrit32:
Working through some edge cases related to no-MANE and UTR regions

theferrit32:
[...]

PR #437, opened 2026-04-24
Genomic to protein projection

theferrit32:
Touch #146

This PR adds:
- forward transcription of a genomic variant to a transcript variant (preferring MANE using cool-seq-tool's logic)
- forward translation of a transcript variant to a protein variant
- automatic registration of a transcript and protein form of a genomic variant in `PUT /variation`
- automatic registration of a protein form of a transcript variant in `PUT /variation`


src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1083 @@
...
+        directly to the associated protein.
+
+        This method catches and suppresses major error cases and communicates
+        results as warning messages.
+
+        TODO: This only supports refseq accessions. Consider another approach.
+
+        :param variation: variation to project
+        :param storage: Storage instance
+        :return: list of warning messages, or None if completely successful
```
jsstevenson: would we be able to just raise exceptions? this seems like a counterintuitive response model (and returning specific exceptions makes for more ergonomic error handling at the next level up)

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1083 @@
...
+            location.start,
+            location.end,
+            lambda: _get_transcript_literal_state(self.dp, variation),
+            missing_protein_message=(
+                "Projection skipped: no associated protein accession for transcript "
+                f"{transcript_ac}"
+            ),
+        )
+
+    def add_mappings(
```
jsstevenson: small quibble but i think this should have a more precise name

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1083 @@
...
+    start: int
+    end: int
+    state: models.LiteralSequenceExpression
+
+
+def _get_variation_location(
+    variation: VrsVariation,
+    *,
+    require_refget: bool = False,
+) -> tuple[_ProjectionLocation | None, list[str] | None]:
```
jsstevenson: if these two elements are mutually exclusive, I don't think we should be returning a tuple where each part is nullable. Ideally this would just be another raised exception I think

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1083 @@
...
+    except AttributeError:
+        return None, ["Projection unsupported: variant lacks sequence location details"]
+
+    if not isinstance(start, int) or not isinstance(end, int):
+        return None, ["Projection unsupported for variants with Range positions"]
+
+    return _ProjectionLocation(start, end, refget_accession), None
+
+
+def _get_refseq_accession(dp: _DataProxy, refget_accession: str) -> str | None:
```
jsstevenson: Maybe I'm crazy but I feel like this is already implemented somewhere, either in coolseqtool or somewhere in vrs-python/anyvar (or both?)

jsstevenson (changes requested):
some initial thoughts

src/anyvar/mapping/projection.py
theferrit32: Unfortunately I don't think it is. `translate_sequence_identifier(ga4gh_id, "refseq")` is there but there's no existing code to select an appropriate value if it returns multiple.

src/anyvar/mapping/projection.py
theferrit32: done

src/anyvar/mapping/projection.py
theferrit32: I changed a bunch of paths to now raise exceptions instead of returning optional error messages

src/anyvar/mapping/projection.py
jsstevenson: Yup you're totally right 👍

jsstevenson (commented):
I'm getting weird github server errors trying to leave a review but

* add env var to .env config
* add a blurb to features.rst about auto-add of projection stuff during registration
* I think the `UTA_DB_URL` check already happens in `CoolSeqTool` initialization so doesn't need to be handled by AnyVar

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1049 @@
...
+        if not isinstance(variation, models.Allele):
+            raise ProjectionError(
+                "Projection unsupported: only Allele variations are supported"
+            )
+
+        location = _get_variation_location(variation, require_refget=True)
+        if location.refget_accession is None:
+            raise ProjectionError(
+                "Projection unsupported: variant lacks sequence location details"
+            )
```
jennifer-bowser: Minor comment - would it make sense to move this check into '_get_variation_location`?

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1049 @@
...
+
+        try:
+            refseq_accession = _refget_to_refseq_accession(
+                self.dp, location.refget_accession
+            )
+            if not refseq_accession:
+                return
```
jennifer-bowser, line 1074: Should we log something here if we can't project the variant due to being unable to get a refseq accession from the refget one?

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1049 @@
...
+            raise ProjectionError(msg)
+
+        input_vrs_id: str = variation.id  # type: ignore
+
+        # Get the RefSeq genomic accession (NC_xxx)
+        if alt_ac is None:
+            alt_ac = _refget_to_refseq_accession(self.dp, location.refget_accession)
+        if not alt_ac or not alt_ac.startswith("NC_"):
+            _logger.debug("Skipping projection: %s is not a genomic accession", alt_ac)
+            return  # silently skip non-genomic variants
```
jennifer-bowser: I could be misunderstanding, but I think we won't ever hit this code block? 

The flow looks to be `add_projections -> _add_projections_for_refseq_accession -> _project_genomic_variant`. We first compute the refseq accession in `add_projections`, and we return early if it's `None` at that point. Then in `_add_projections_for_refseq_accession`, we check to see if the accession starts with `"NC"` before calling `_project_genomic_variant`. So I think we're already handling both of these checks upstream!

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1049 @@
...
+            self._loop.call_soon_threadsafe(self._loop.stop)
+            self._thread.join(timeout=timeout)
+
+            if self._thread.is_alive():
+                _logger.warning("Projector event loop thread did not stop cleanly")
+                return
+
+        self._loop.close()
+
+    def _run_async_projection(
```
jennifer-bowser, line 658: Ooh good call making this run asynchronously!

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1049 @@
...
+            alt_ac,
+            location.start,
+            location.end,
+        )
+
+        # Use cool-seq-tool to get MANE c./p. representations, falling back to
+        # the longest compatible remaining transcript when MANE is unavailable
+        # or incompatible.
+        result = self._run_async_projection(
+            self.cst.mane_transcript.grch38_to_mane_c_p(
```
jennifer-bowser: It looks like this function is GRCh38-specific. What happens if someone registers a GRCh37 variant?

src/anyvar/mapping/projection.py
```diff
@@ -0,0 +1,1049 @@
...
+        """
+        self.cst = cst
+        self.dp = dp
+        self._loop = asyncio.new_event_loop()
+        self._thread = threading.Thread(
+            target=self._loop.run_forever, daemon=True, name="projector-loop"
+        )
+        self._thread.start()
+
+    def close(self, timeout: float = 5.0) -> None:
```
jennifer-bowser, line 634: Codex has informed me that this function stops and closes the event loop here without closing CoolSeqTool/UTA’s asyncpg connection pool? 

It says:
> After the first successful projection, UTA opens its pool lazily on this loop; closing the loop directly can leave DB connections/tasks dangling on app shutdown or reload. I’d add a coroutine scheduled on self._loop before loop.stop() to close self.cst.uta_db._connection_pool when present.

I'm not 100% sure on whether or not this is accurate; just thought I'd throw it out there.

jennifer-bowser (changes requested):

src/anyvar/mapping/projection.py
theferrit32: done

src/anyvar/mapping/projection.py
theferrit32: done

src/anyvar/mapping/projection.py
theferrit32: Agree, I simplified it to assume the genomic_ac is what was passed in

src/anyvar/mapping/projection.py
theferrit32: I'm thinking through this. I think we don't want to have two paths of lifting over GRCh37 variants in our application. Our REST API layer can/does already do liftover for GRCh37 inputs. So I think we should use that lifted over GRCh38 version for the input to the projector, and make the VariantProjector class *require* genomic inputs to be GRCh38. At least for now.

theferrit32: Revised my above thoughts. We'll support GRCh37 inputs to VariantProjector and let cool-seq-tool attempt to project forward to GRCh38 in order to try to get to a MANE transcript, even if AnyVar's logic around the unambiguous round-trippability decided not to persist a GRCh38 form.

src/anyvar/mapping/projection.py
theferrit32: I'm not sure how important this is since it's a long-lived process and we don't create/destroy multiple VariantProjector objects over the lifespan of the process. But might help if there's a SQL query in progress, and let it finish before exiting the process. I'll add a call to close the `_connection_pool`.

jsstevenson (dismissed):

compose.yaml
```diff
@@ -78,6 +78,7 @@ services:
       - SEQREPO_DATAPROXY_URI=seqrepo+file:///usr/local/share/seqrepo/2024-12-20
       - UTA_DB_URL=postgresql://anonymous:anonymous@uta:5432/uta/uta_20241220
       - ANYVAR_STORAGE_URI=postgresql://anyvar:anyvar-pw@anyvar_db:5432/anyvar
+      - ANYVAR_ENABLE_PROJECTION=true
```
jennifer-bowser, line 81: It looks like `Dockerfile` installs `.[dev,test,queueing,postgres]` and omits `projection` (see line 24), so I think projection still won't work yet out-of-the-box

jennifer-bowser (changes requested):
Looks really good, just one last minor change and then I think this is good to go!

compose.yaml
theferrit32: Good catch, I just added it to the dockerfile

https://github.com/biocommons/anyvar/pull/437/changes/d2dda160908ee9b4ad134adf67ba9e6d08ce2b68

jennifer-bowser (approved):

PR #440, opened 2026-04-28
feat!: rename Annotations to Extensions

### LINKED ISSUE #438: Rename "Annotations" to "Extenstions"
In our DevTime meeting on 4/24, we decided to rename "Annotations" to "Extensions". "Annotations" suggests a use case in line with VA-Spec Annotations, when this table is actually intended to support the more general use case of a VA-Spec [Extension](https://vrs.ga4gh.org/en/latest/concepts/AdditionalDataTypes/index.html#extension). 

This will involve:
- Renaming the table
- Updating all associated class names
- Updating all associated function names
- Documentation - see below

Re: Documentation - we should give examples of the sorts of information this is meant for, and the sorts of information this is **not** meant for, including the technical limitations at play. I am unsure of the best place to put this info, though.

jsstevenson:
close #438 

* Update all code + lengthen the description in docs to clarify intent. Hopefully I caught everything!
* Also includes some minor name tweaking for related properties. In particular, I changed "annotation_type" to "extension_name" (or just "name") to match the VRS object pattern. 
* Did some other cleanup on stuff that had been changed but not updated in the docs
* You can see the docs changes here: https://anyvar--440.org.readthedocs.build/en/440/features.html#object-extensions 

I actually think this is a very good change, we previously had a lot of mentions of "VCF annotations" that could've been easy to confuse with the DB annotation object.

### misc

* Would be a great time to have a migration manager!
* I wonder if object GET endpoints should optionally be able to return extensions in the objects themselves


jennifer-bowser (approved):
In `tests/unit/storage/storage_test_funcs.py` there are a few variable names and comments that still refer to "annotations" instead of "extensions"; but since that doesn't affect functionality I'm approving anyways!

PR #459, opened 2026-05-11
feat: delete extensions

### LINKED ISSUE #425: Enable deletion of annotations
### Feature description

Currently, users can only add annotations; but they need to be able to delete them as well. We should add a new HTTP endpoint to support this + underlying Storage class methods. Don't forget to write tests as well!

### Use case

As a user, I want to be able to delete annotations that I no longer need or that were entered in error.

### Acceptance Criteria

Annotations can be deleted via HTTP endpoint.

jsstevenson:
close #425 

replace existing storage class extension deletion method with one that takes an object ID and optionally an extension name(, and optionally a value), and performs a deletion specific to the parameters given. A rowcount is returned so that a client can figure out whether a delete did nothing because no extension matched those parameters or if it was because no object matching that ID exists.

Somewhat related, but writing it out: currently, extensions can be duplicative for object ID-name-value. This is because it's annoying to enforce uniqueness on anything that includes a JSON value.

### DIFF
--- a/src/anyvar/anyvar.py
+++ b/src/anyvar/anyvar.py
@@ -213,7 +213,7 @@ class AnyVar:
         except KeyError as e:
             raise ObjectNotFoundError from e
         for extension in self.object_store.get_extensions(object_id):
-            self.object_store.delete_extension(extension)
+            self.object_store.delete_extensions(vrs_object.id, extension.name)
         for mapping in self.object_store.get_mappings(object_id, as_source=True):
             self.object_store.delete_mapping(mapping)
         for mapping in self.object_store.get_mappings(object_id, as_source=False):
@@ -242,6 +242,29 @@ class AnyVar:
             except KeyError as e:
                 raise ObjectNotFoundError(object_id) from e
         return extensions
+
+    def delete_object_extensions(
+        self, object_id: str, extension_name: str | None = None
+    ) -> None:
+        """Delete extensions
+
+         * If no ``extension_name`` is given, delete all associated extensions.
+         * Delete all instances of extensions by the given name
+
+        :param object_id: object ID to delete extensions for
+        :param extension_name: optionally, the name of extensions to delete
+        :raise ObjectNotFoundError: if ``object_id`` can't be found in DB
+        """
+        try:
+            rowcount = self.object_store.delete_extensions(object_id, extension_name)
+        except Exception as e:
+            _logger.exception("Failed to retrieve extensions for object: %s", object_id)
+            raise e  # noqa: TRY201
+        if not rowcount:
+            try:
+                _ = self.get_object(object_id)
+            except KeyError as e:
+                raise ObjectNotFoundError(object_id) from e
 
     def create_timestamp_if_missing(self, object_id: str) -> int | None:
         """Store a 'creation_timestamp' extension if missing for an object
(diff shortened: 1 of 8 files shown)


src/anyvar/storage/postgres.py
```diff
@@ -347,6 +348,24 @@ def delete_extension(self, extension: metadata.Extension) -> None:
         with self.session_factory() as session, session.begin():
             session.execute(stmt)
 
+    def delete_extensions(
```
jennifer-bowser, line 338: Do we need both a singular and a plural deletion function for extensions? I feel it'd be less maintenance work going forward to just have one function that can do both; but if others disagree we can certainly keep both!

jsstevenson: `delete_extension` deletes everything that matches a specific `name` and `value` (which, tbf, could end up deleting multiple extensions!), but `delete_extensions` just deletes everything matching the `name`. I think these are distinct things but also we maybe don't need to keep both around. I'm not sure which one is more useful. It's been a week or two since I worked on this so my recollection is imperfect, but I think I was trying to avoid doing anything API-breaking. But if we feel good about this, we could also just delete `delete_extension`

jsstevenson:
btw some tests are now failing and lint isn't passing so this is going to draft again for a sec

src/anyvar/storage/postgres.py
jennifer-bowser: How about if we combine them so that it's one function that takes an optional key - if key is provided, delete just those extensions; else delete all.

src/anyvar/anyvar.py
```diff
@@ -200,6 +200,18 @@ def put_extension(self, extension: metadata.Extension) -> int | None:
             raise e  # noqa: TRY201
         return extension_id
 
+    def delete_extensions(self, vrs_id: str, extension_name: str) -> None:
```
jennifer-bowser: It looks like this function isn't used anywhere? Is this one we can get rid of in favor of `delete_object_extensions` (on line 258)?

tests/integration/restapi/objects_router/test_extension.py
jennifer-bowser, line 17: It looks like we're not using `preloaded_alleles` anymore. Can we nuke it?

jennifer-bowser: Oh wait we stopped using that a while ago it seems. Never mind, we can handle that in a later PR

src/anyvar/anyvar.py
jsstevenson: Ha this must've been a brainfart on my part

tests/integration/restapi/objects_router/test_extension.py
jsstevenson: Importing the `preloaded alleles` fixture loads data into the storage instance in `restapi_client`. I think this is a little obtuse though, I'm not sure why it had to be done this way

jsstevenson: I think it's because some methods want access to the data in those alleles, which is why they're passed as a separate object, but there are occasional cases where you want the preloaded side-effect but you don't need the variation object itself. I added some explanation here

jennifer-bowser (approved):

Issue #340, opened 2025-11-25
Add back copy number liftover success case to test


korikuzma:
https://github.com/biocommons/anyvar/blob/52231711068daf192a33705cb606ccd71b87f686/tests/unit/extras/test_liftover.py#L12-L25

https://github.com/biocommons/anyvar/blob/52231711068daf192a33705cb606ccd71b87f686/tests/unit/extras/test_liftover.py#L127-L133

Originally added by @jsstevenson

PR #461, opened 2026-05-14
refactor: fix type errors (issue #259)

jennifer-bowser:
This PR fixes some type errors that I fixed as I attempted to get Pyright added to our codebase. Pyright is no longer being added, but we might as well ship the type improvements I made.

Also, I got rid of all copy number/copy change related code, since that confused type-checkers a lot and it feels weird to only half-support these variants anyways. I documented what I deleted in issue #340 for when we're ready to re-enable support for these.


src/anyvar/anyvar.py
```diff
@@ -131,8 +134,10 @@ def get_object(
         """
         if object_type is not None:
             # Search specific object type
-            found = self.object_store.get_objects(
-                object_type=object_type, object_ids=[object_id]
+            found: list[SupportedVrsObject] = list(
```
jsstevenson, line 147: fwiw this could just be `objects.SupportedVrsObject`

jsstevenson (dismissed):

jsstevenson (approved):
🚀

PR #470, opened 2026-05-28
feat: register protein sequence consequence/canonical allele catvars

### LINKED ISSUE #453: Build registration endpoint for `ProteinSequenceConsequence` and `CanonicalAllele` categorical variations
### Feature description

We need to support categorical variants in AnyVar. That means we need to build an endpoint to allow for CatVar registration. We're starting by supporting `ProteinSequenceConsequence` and `CanonicalAllele` CatVars with only one `DefiningAlleleConstraint`. See Requirement #447 for details.
 
### Use case

I'm frustrated when a tool proclaiming to support "Any Variant" does not, in fact, support "any" variant.

### Acceptance Criteria

See Requirement #447 

### Proposed solution

* A new endpoint called `PUT /categorical_variant` or something similar, or else a CLI load command.
* Probably need a new table ~~or tables (one per kind of catvar?)~~
    * We'll use one table that supports CatVars w/ a single DefiningAlleleConstraint. In the future, other kinds of CatVars can be stored in other tables. 
* The property indicating what kind of CatVar this is ~~can either be an extension/other field on the CatVar itself,~~ will be an external field

### Alternatives considered

_No response_

### Implementation details

- When a CanonicalAllele is registered, we'll do all of the mappings that we normally do for variation registration (i.e., transcription, translation, liftover)
- We assume that the DefiningAlleleConstraint for Canonical Alleles is a genomic variant and for ProteinSequenceConsequence it will be a protein variant. **AnyVar should validate this though.** => If this is complicated, just write a stub function and we'll come back to this in another ticket.

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

None

jsstevenson:
close #453

## Decisions

* `/categorical_variants` not `/categorical_variations`. I don't even know which is correct at this point
* the "reference" to the external resource is just the ID of the catvar itself. So rather than making a catvar with an "xref" to a clingen allele record, the `id` field is just `clingen.allele:CA12345`. And similarly, fetching that specific catvar is `GET /categorical_variants/canonical_alleles/clingen.allele:CA12345`
* Rather than a general "drop in a catvar, have some indication of its type, let anyvar figure it out" endpoint, I just made resource type-specific registration endpoints, I think that's better for our sanity and avoids inventing any new conventions. Similarly, there are separate sql tables for each categorical variant type. I had wanted to be slick but it probably makes more sense to just use a new table for each type.
* Make new classes for PSQ's/CA's instead of reusing `ga4gh.cat_vrs` recipes, a couple reasons
    1) we want to be stricter about some validation things (ID is required, `len(constraints) == 1`)
    2) we might want to interpret relations differently than the recipes do, now or in the future
* Validating molecule type is done outside of the pydantic `field_validator` because it might require seqrepo access. Alternatives here could be 1) requiring the submitter to explicitly declare molecule type and trusting it, or 2) injecting validation context into the fastapi route validation logic, but that looked like it was going to be insane.  It's also possible that this function will duplicate work from the projection end point so whichever one gets merged second can resolve that.
* Request paths below -- also add basic GET by ID for round-trip testing and also because it just seems reasonable and enabled roundtrip testing
* Haven't added any `relations` to the input or output objects yet, pending final determination on how we implement our search functions. But broadly, I think you could just make sure to add in the right values when the object is coming back out to the user. Not sure if it's worth validating them on the way in. 
* Haven't updated docs yet, pending final search endpoint behavior
* Didn't add a Snowflake implementation because a) scary and b) maybe we are ending snowflake support?

## Routes

`PUT /categorical_variants/canonical_alleles` -- to register a CA
`GET /categorical_variants/canonical_alleles/<id>` -- to fetch a CA by ID
`PUT /categorical_variants/protein_sequence_consequences` -- to register a PSQ
`GET /categorical_variants/protein_sequence_consequences/<id>` -- to fetch a PSQ by ID

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
     "python-multipart",  # required for fastapi file uploads
     "uvicorn",
     "ga4gh.vrs[extras]>=2.3.2,<3.0",
+    "ga4gh.cat_vrs>=0.7.1",
     "sqlalchemy~=2.0.0",
     "snowflake-sqlalchemy~=1.8.2",
     "pyyaml",
--- a/src/anyvar/core/categorical_variants.py
+++ b/src/anyvar/core/categorical_variants.py
@@ -1,0 +1,108 @@
+"""Provide core typing/structure for categorical variants."""
+
+import logging
+
+from ga4gh.cat_vrs import models as cat_vrs_models
+from ga4gh.vrs import models as vrs_models
+from ga4gh.vrs.dataproxy import _DataProxy
+from pydantic import Field, field_validator
+
+_logger = logging.getLogger(__name__)
+
+
+def is_expected_molecule_type(
+    seq_ref: vrs_models.SequenceReference,
+    expected_type: vrs_models.MoleculeType,
+    dp: _DataProxy,
+) -> bool:
+    """Validate that the given sequence reference has the expected molecule type
+
+    Currently only intended to support genomic and protein variants since that's what
+    we have catvars for
+
+    No further validation is performed if the `moleculeType` property is defined; its
+    value is accepted as authoritative and no additional sequence type lookup is
+    performed.
+
+    :param seq_ref: defining allele context's sequence reference object
+    :param expected_type: desired molecule type
+    :param dp: proxy for seq alias lookup
+    :return: whether the sequence reference's molecule type can be validated as consistent
+        with the expected molecule type
+    """
+    if seq_ref.moleculeType:
+        return seq_ref.moleculeType == expected_type
+
+    refseq_aliases = dp.translate_sequence_identifier(
+        f"ga4gh:{seq_ref.refgetAccession}", "refseq"
+    )
+    if not refseq_aliases:
+        _logger.debug(
+            "Unable to translate sequence accession %s to a known namespace",
+            seq_ref.refgetAccession,
+        )
+        return False
+    raw_seq_type = dp.extract_sequence_type(refseq_aliases[0])
+    if expected_type == vrs_models.MoleculeType.GENOMIC:
+        return raw_seq_type == "g"
+    if expected_type == vrs_models.MoleculeType.PROTEIN:
+        return raw_seq_type == "p"
+    _logger.debug(
+        "Encountered unexpected or unknown sequence type for %s: %s",
+        seq_ref.refgetAccession,
+        raw_seq_type,
+    )
+    return False
+
+
+class _SimpleAlleleCatVar(cat_vrs_models.CategoricalVariant):
+    """Abstract class for a simple Defining Allele Constraint-based catvar"""
+
+    id: str  # type: ignore[reportIncompatibleVariableOverride]
+    constraints: list[cat_vrs_models.Constraint] = Field(min_length=1, max_length=1)  # type: ignore[reportIncompatibleVariableOverride]
+
+    @field_validator("constraints")
+    @classmethod
+    def validate_constraints(
+        cls, constraints: list[cat_vrs_models.Constraint]
+    ) -> list[cat_vrs_models.Constraint]:
+        """Validate constraints property
+
+        :param constraints: Constraints property to validate
+        :return: Constraints property
+        :raises ValueError: If constraints property does not satisfy the requirements
+        """
+        constraint = constraints[0]  # guaranteed by pydantic to have exactly 1 element
+        if not isinstance(constraint.root, cat_vrs_models.DefiningAlleleConstraint):
+            raise ValueError("Constraint must be a DefiningAlleleConstraint")  # noqa: TRY004
+        return constraints
+
+
+class ProteinSequenceConsequence(_SimpleAlleleCatVar):
+    """ProteinSequenceConsequence catvar
+
+    Validates *shape* of input data, but does not validate *values*, ie that
+    given defining allele is of the expected molecule type, because that
+    requires external data access to SeqRepo (and that's messy to inject in a FastAPI endpoint).
+
+    The validations performed do differ from the original Cat-VRS recipes, which is why we
+    have a separate class.
+
+    It doesn't need any additional code implemented beyond its parent class,
+    but we still give it a separate class for mapping purposes in the ORM
+    """
+
+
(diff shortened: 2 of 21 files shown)


src/anyvar/core/categorical_variants.py
```diff
@@ -0,0 +1,104 @@
...
+
+class _SimpleAlleleCatVar(cat_vrs_models.CategoricalVariant):
+    """Abstract class for a simple Defining Allele Constraint-based catvar"""
+
+    id: str  # type: ignore[reportIncompatibleVariableOverride]
+    constraints: list[cat_vrs_models.Constraint] = Field(min_length=1, max_length=1)  # type: ignore[reportIncompatibleVariableOverride]
+
+    @field_validator("constraints")
+    @classmethod
+    def validate_constraints(
```
jennifer-bowser, line 66: Is it possible to have more than one defining allele? If so, should we be checking here to ensure there's only one? Or does that not matter?

src/anyvar/core/categorical_variants.py
```diff
@@ -0,0 +1,104 @@
...
+        cls, v: list[cat_vrs_models.Constraint]
```
jennifer-bowser: This is subjective, and perhaps a bit pedantic; but I prefer to use longer, more descriptive variable names vs abbreviated ones. Could we consider updating `v` to something more semantic?

tests/integration/restapi/vcf_router/test_ingest_vcf.py
```diff
@@ -50,7 +50,7 @@ def basic_vcf():
 ##INFO=<ID=VRS_Error,Number=.,Type=String,Description="If an error occurred computing a VRS Identifier, the error message">
 ##INFO=<ID=VRS_Starts,Number=R,Type=Integer,Description="Interresidue coordinates used as the location starts for the GA4GH VRS Alleles corresponding to the GT indexes of the REF and ALT alleles">
 ##INFO=<ID=VRS_Ends,Number=R,Type=Integer,Description="Interresidue coordinates used as the location ends for the GA4GH VRS Alleles corresponding to the GT indexes of the REF and ALT alleles">
-##INFO=<ID=VRS_States,Number=R,Type=String,Description="The literal sequence states used for the GA4GH VRS Alleles corresponding to the GT indexes of the REF and ALT alleles">
+##INFO=<ID=VRS_States,Number=R,Type=String,Description="The literal sequence states used for the GA5GH VRS Alleles corresponding to the GT indexes of the REF and ALT alleles">
```
jennifer-bowser: Was this an intentional change?

jsstevenson: nope, thought I fixed this 😢

src/anyvar/core/categorical_variants.py
jsstevenson: that's checked a few lines up:

```
constraints: list[cat_vrs_models.Constraint] = Field(min_length=1, max_length=1)
```

src/anyvar/core/categorical_variants.py
jsstevenson: This is the typical pydantic validator pattern, but probably fine to rename it here

jsstevenson:
pulled in latest commits to main + added more tests

src/anyvar/restapi/categorical_variants_router.py
```diff
@@ -0,0 +1,169 @@
...
+            root=cat_vrs.DefiningAlleleConstraint(
+                allele=vrs.Allele(
+                    id="ga4gh:VA.j4XnsLZcdzDIYa5pvvXM7t1wn9OITr0L",
+                    digest="j4XnsLZcdzDIYa5pvvXM7t1wn9OITr0L",
+                    location=vrs.SequenceLocation(
+                        id="ga4gh:SL.t-3DrWALhgLdXHsupI-e-M00aL3HgK3y",
+                        digest="t-3DrWALhgLdXHsupI-e-M00aL3HgK3y",
+                        sequenceReference=vrs.SequenceReference(
+                            refgetAccession="SQ.cQvw4UsHHRRlogxbWCB8W-mKD4AraM9y",
+                        ),
```
theferrit32, line 42: Can we add a `moleculeType="protein"` or `moleculeType=vrs.MoleculeType.PROTEIN` here?

src/anyvar/storage/sqlalchemy.py
```diff
@@ -447,3 +450,61 @@ def search_alleles(
...
+        This method **is not** responsible for validating that the provided catvar meets
+        data requirements to be considered a Canonical Allele instance;
+        passing an object without prior validation may raise unexpected errors
+
+        :param ca: canonical allele catvar
+        """
+        db_entity: orm.CanonicalAllele = mapper_registry.to_db_entity(ca)
+        with self.session_factory() as session, session.begin():
+            session.merge(db_entity.allele)
+            session.flush()
```
theferrit32: I'm not sure if these `session.flush()` calls are needed, but probably no harm

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "python-multipart",  # required for fastapi file uploads
     "uvicorn",
     "ga4gh.vrs[extras]>=2.3.2,<3.0",
+    "ga4gh.cat_vrs",
```
theferrit32: Should the version maybe be pinned or constrained to a range?

src/anyvar/storage/sqlalchemy.py
jsstevenson: Hm. I believe I put this in because something hadn't been working and it seemed to fix it. But now tests pass without the `flush` so... might as well just pull it out and see what happenes

src/anyvar/restapi/categorical_variants_router.py
jsstevenson: Yeah. I meant to do this before, but also updated the molecule type validation function to trust `moleculeType` blindly.

theferrit32 (approved):

src/anyvar/restapi/categorical_variants_router.py
theferrit32: We could add more around this if we do https://github.com/biocommons/anyvar/discussions/478

PR #472, opened 2026-05-29
Allow variation registration in stateless mode

### LINKED ISSUE #471: `PUT /variation` fails with `NoObjectStore`
### Describe the bug

Variation registration fails when using the `NoObjectStore`

### Steps to reproduce

1. Run AnyVar with `NoObjectStore` - `ANYVAR_STORAGE_URI=""`
2. Attempt to register an allele.  For example:
```shell
echo '{"definition":"22-43928847-C-G", "input_type":"Allele"}' | curl -XPUT -d @- -H "Content-type: application/json" "http://localhost:8000/variation"
```

### Expected behavior

200 Response with VRS object as the body

### Current behavior

Internal Server Error
```
 File "./anyvar/src/anyvar/restapi/objects_router.py", line 206, in register_variation
    responses: list[RegisterVariationResponse] = _register_variations(av, [variation])
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./anyvar/src/anyvar/restapi/objects_router.py", line 169, in _register_variations
    av.create_timestamp_if_missing(translation_result.variation.id)  # type: ignore (ID guaranteed to be present)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./anyvar/src/anyvar/anyvar.py", line 252, in create_timestamp_if_missing
    timestamps: list[metadata.Extension] = self.get_object_extensions(
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./anyvar/src/anyvar/anyvar.py", line 242, in get_object_extensions
    raise ObjectNotFoundError(object_id) from e
```

### Acceptance Criteria

Variation registration succeeds with `NoObjectStore`

### Possible reason(s)

`AnyVar.get_object_extensions()` attempts to fetch a VRS object from the store when no extensions are found. 

### Suggested fix

_No response_

### Branch, commit, and/or version

branch: main
commit: 1d23f4932ca4da923f9381eb709361ebe54d7836

### Screenshots

_No response_

### Environment details

MacBook M4
Python 3.12

### Additional details

_No response_

### Contribution

None

ehclark:
Variation registration calls `create_timestamp_if_missing()` to populate the timestamp extension for the registered variation.  This in turn calls `get_object_extensions()` to determine if the extension already exists.  `get_object_extensions()` throws an error if the variation is not found.  In stateless mode, the `NoObjectStore` always throws an error on `get_object()`.

Changed code to not check for variation existence in store when in stateless mode.

### DIFF
--- a/src/anyvar/anyvar.py
+++ b/src/anyvar/anyvar.py
@@ -17,6 +17,7 @@ from anyvar.core import metadata, objects
 from anyvar.core.objects import SupportedVrsObject
 from anyvar.storage import DEFAULT_STORAGE_URI
 from anyvar.storage.base import Storage
+from anyvar.storage.no_db import NoObjectStore
 from anyvar.translate.base import Translator
 from anyvar.translate.vrs_python import VrsPythonTranslator
 
@@ -235,7 +236,7 @@ class AnyVar:
         except Exception as e:
             _logger.exception("Failed to retrieve extensions for object: %s", object_id)
             raise e  # noqa: TRY201
-        if not extensions:
+        if not extensions and not isinstance(self.object_store, NoObjectStore):
             try:
                 _ = self.get_object(object_id)
             except KeyError as e:


jsstevenson (approved):
I think this is fine as a quick fix, but this feels like an indication of a deeper design problem

alternatives would be to catch the `ObjectNotFoundException` in the route function or in the timestamp creation function, but I'm not sure either is better

PR #477, opened 2026-06-02
feat!: add DuckDB storage backend

### LINKED ISSUE #475: Add back DuckDb storage backend
Experimental!

Try out a DuckDb backend with a single file for storage. Potentially usable for GKS-AnVIL efforts.

jsstevenson:
close #475 

* Refactor most of the Postgres class into a `SqlAlchemyBase` class, to be reused by both Postgres and DuckDB. There is a tiny amount of postgres-specific code to maintain but most basic schema alterations can happen in the sqlalchemy base class, hopefully.
* Alter schema for mappings so that it enforces against row-duplication instead of using an autoincrementing ID key. This does technically change behavior but I think it's fine -- duplicate inserts are just idempotent instead of creating redundant rows (and redundancy is probably actively bad in the case of a mapping, for example). I think I could revert this though.
* Add DuckDB storage implementation. As far as I can tell it... totally works?  And is pretty straightforward once we've done the above. Add basic docs about usage. There are tests, but as with Snowflake, you can't easily run tests for different SqlAlchemy engines in one test run because of insane global state things, so it mirrors the way you select for different DB engines. This might be worth investigating later.

## Questions

* How to run DuckDB in CI/CD? We could easily parametrize the tests by storage URI so they run for the intersection of postgres/duckdb and 3.11/3.12/3.13. Is that overkill? Could also just run the duckdb-specific unit tests in a separate run.
    * Answer: to be addressed in #482

### DIFF
--- a/docs/source/api/api/storage/anyvar.storage.duckdb.rst
+++ b/docs/source/api/api/storage/anyvar.storage.duckdb.rst
@@ -1,0 +1,9 @@
+﻿anyvar.storage.duckdb
+=====================
+
+.. automodule:: anyvar.storage.duckdb
+   :members:
+   :no-inherited-members:
+   :undoc-members:
+   :special-members: __init__
+   :exclude-members: model_fields, model_config, metadata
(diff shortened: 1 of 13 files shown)


jsstevenson:
this buddy is back 2 draft for a minute to merge in latest changes

jennifer-bowser (approved):

PR #480, opened 2026-06-04
Async Request-Response for Bulk Variation Registration

### LINKED ISSUE #473: Add async request-response (queueing) for `PUT /variations` endpoint
### Feature description

The `/vcf` endpoints provide optional for asynchronous request-response pattern using Celery as a queueing and worker management backend.  Support the same asynchronous request-response pattern for the `/variations` endpoint.

### Use case

The asynchronous request-response pattern allow for better reliability and scalability in cloud environments because it supports long running requests (>30 secs) without having to maintain open HTTP connections.  The queueing and worker support provided by Celery provides better horizontal scalability and adaptability in shared compute environments.

### Acceptance Criteria

A `run_async` query parameter can be provided to the `/variations` endpoint to process the request using the asynchronous request-response pattern.  When `run_async` is `true`, a `run_id` can be optionally provided to identify the processing job during the response polling phase.  If `run_id` is not provided, a random identifier will be generated.

If queueing is not enabled, passing `run_async=true` results in a 400 response.

### Proposed solution

Implement asynchronous request-response processing for `/variations` requests using the same pattern established for the `/vcf` endpoints.

### Alternatives considered

_No response_

### Implementation details

_No response_

### Potential Impact

The changes will be fully backwards compatible with existing functionality.

### Additional context

_No response_

### Contribution

None

ehclark:
### Summary

This PR adds asynchronous request-response support to the `PUT /variations` endpoint, following the same [async request-reply pattern](https://learn.microsoft.com/en-us/azure/architecture/patterns/async-request-reply) already used by the `/vcf` endpoints. Large batches of variation registrations can now be submitted asynchronously via `PUT /variations?run_async=true`, returning a `202 Accepted` with a `run_id` that the client polls at `GET /variations/{run_id}` until the result is ready.

As part of this work, the shared async task lifecycle logic (Celery state machine handling, run-ID validation, async-enabled checks) was extracted from the VCF router into a new reusable module, and the VCF router was refactored to consume it.

### Key Changes

**New module: `src/anyvar/restapi/async_utils.py`**
Extracts three reusable helpers previously inline in the VCF router:
- `check_async_enabled()` — returns an `ErrorResponse` (400) if async prerequisites are missing
- `validate_run_id_available()` — rejects duplicate `run_id` values that are still active in Celery
- `resolve_async_task_status()` — encapsulates the full Celery task state machine: SUCCESS → call `on_success` callback and forget; FAILURE → extract error code, optional cleanup, forget; PENDING → poll up to 5s then 404; SENT → 202 with `Retry-After`

**Modified: `src/anyvar/restapi/variations_router.py`**
- `PUT /variations` gains `run_async` (bool) and `run_id` (optional str) query parameters. When `run_async=true`, the request body is serialized and dispatched to a Celery task, returning a `RunStatusResponse` with `Location` and `Retry-After` headers.
- New `GET /variations/{run_id}` endpoint delegates to `resolve_async_task_status()` to return status or completed results.

**Modified: `src/anyvar/restapi/vcf_router.py`**
Refactored to import and use the shared helpers from `async_utils` instead of inline implementations.

**New Celery task: `register_variations` in `src/anyvar/queueing/celery_worker.py`**
Deserializes `VariationRequest` dicts, calls `_register_variations()` via the worker's `AnyVar` instance, and returns serialized `RegisterVariationResponse` dicts. Uses the same `enter_task()`/`exit_task()`/`maybe_teardown_anyvar_app()` lifecycle as existing VCF tasks.

**New: `has_variations_queueing_enabled()` in `src/anyvar/anyvar.py`**
A lighter prerequisite check than the VCF equivalent — requires only `celery` + `CELERY_BROKER_URL` (no `aiofiles` or `ANYVAR_VCF_ASYNC_WORK_DIR`), since variation registration doesn't involve file I/O.

### Configuration

New environment variables (documented in `docs/source/configuration/async.rst`):

| Variable | Default | Purpose |
|---|---|---|
| `ANYVAR_VARIATIONS_ASYNC_FAILURE_STATUS_CODE` | `500` | HTTP status returned by `GET /variations/{run_id}` on task failure |
| `ANYVAR_EXPECTED_VARIATIONS_PER_SECOND` | `100` | Used to estimate `Retry-After` header value |

Async variation registration requires only `CELERY_BROKER_URL` and `CELERY_BACKEND_URL` — not the VCF-specific `ANYVAR_VCF_ASYNC_WORK_DIR`.

### Tests

- **`tests/unit/restapi/test_async_utils.py`** — unit tests for all three shared helpers covering success, failure, timeout, cleanup, and edge cases
- **`tests/unit/restapi/test_variations_router.py`** — endpoint tests for `PUT /variations` (async disabled, duplicate run_id, async success, custom run_id, sync fallback) and `GET /variations/{run_id}` (async disabled, delegation to resolver)

### Documentation

- `docs/source/usage/rest_api.rst` — new "Asynchronous Bulk Registration" section with usage examples
- `docs/source/configuration/async.rst` — new "Variation Registration Settings" section; clarifying note that variation async doesn't require `ANYVAR_VCF_ASYNC_WORK_DIR`

**Closes #473**


src/anyvar/restapi/async_utils.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+    return None
+
+
+async def resolve_async_task_status(
+    run_id: str,
+    response: Response,
+    *,
+    on_success: Callable[
+        [AsyncResult],
```
jennifer-bowser: Above we have the try/catch block for importing `queueing` dependencies; but then we use those imports here whether they've been imported or not. So if you're trying to start AnyVar without the queueing dependencies installed, the app breaks when it tries to import this file. 

I asked Codex for solves, and it recommended adding `from __future__ import annotations` - this makes it so that type hints are as strings instead of being evaluated immediately at import time

I also wonder if we should add a `if not _has_async_imports` check to the beginning of this function (similar to the one in the `validate_run_id_available` function)

src/anyvar/anyvar.py
```diff
@@ -94,6 +94,14 @@ def has_queueing_enabled() -> bool:
     )
 
 
+def has_variations_queueing_enabled() -> bool:
+    """Determine whether or not asynchronous task queueing is enabled for variations"""
+    return (
+        importlib.util.find_spec("celery") is not None
+        and os.environ.get("CELERY_BROKER_URL", "") != ""
```
jennifer-bowser, line 108: If I'm tracking correctly here, I believe we also need to check that `CELERY_BACKEND_URL` is set? Or else make that value default to the same as the broker URL in the config setup, instead of defaulting to `None`.

jennifer-bowser (changes requested):
Nice, this looks great! Just a couple of comments for you

src/anyvar/anyvar.py
ehclark: Update config to default `CELERY_BACKEND_URL` to value for `CELERY_BROKER_URL`.

src/anyvar/restapi/async_utils.py
ehclark: `from __future__ import annotations` seems like too blunt of an instrument for my liking.  And it will be deprecated soon anyway.  Updated the code to be sensitive to available imports when defining functions.

jsstevenson: on that note: @ehclark we talked in our group call this morning about whether it would make our lives easier to just incorporate queueing dependencies directly into the main dependency group instead of catching importerrors -- feel free to add thoughts [here](https://github.com/biocommons/anyvar/discussions/483) if you have strong feelings

ehclark:
@jennifer-bowser This is ready for re-review

theferrit32 (approved):

jennifer-bowser (approved):

PR #486, opened 2026-06-16
feat!: make `mapping_type` optional 

### LINKED ISSUE #462: `mapping_type` should be an optional param in the GET mappings endpoint
### Describe the bug

It appears to be required now kbut you should be able to omit it to retrieve all mappings

### Steps to reproduce

its on main

### Expected behavior

above

### Current behavior

bad

### Acceptance Criteria

A user should be able to omit a `mapping_type` parameter and retrieve *all* mappings associated with a variant, not a specific type

### Possible reason(s)

_No response_

### Suggested fix

_No response_

### Branch, commit, and/or version

main

### Screenshots

_No response_

### Environment details

sadflkjsdf

### Additional details

_No response_

### Contribution

None

jsstevenson:
close #462 

I think this was a mistake in how we designed the endpoints -- we want a way to retrieve all mappings, but a path param can't be optional. It also makes more sense to me that `mapping_type` would be like a filter on the `/mappings/` path rather than a subtype. 

So the new path looks like `GET /objects/<vrs_id>/mappings?mapping_type=liftover_to` or `GET /objects/<vrs_id>/mappings`

## Alternatives considered

* Rather than re-engineering this endpoint, just add a new, general `GET /objects/<vrs_id>/mappings` endpoint that retrieves all mappings. I went ahead with just changing the existing endpoint for the reasons above.

### DIFF
--- a/docs/source/usage/rest_api.rst
+++ b/docs/source/usage/rest_api.rst
@@ -69,7 +69,7 @@ The ``/search`` endpoint enables retrieval of all variants that overlap a provid
 Working With Mappings
 =====================
 
-To add a :ref:`mapping <mappings>` between previously-registered variation objects issue a ``PUT`` request to ``/variations/<vrs_id>/mappings``, where ``vrs_id`` is the ``source_id`` of the mapping object:
+To add a :ref:`mapping <mappings>` between previously-registered variation objects issue a ``PUT`` request to ``/object/<vrs_id>/mappings``, where ``vrs_id`` is the ``source_id`` of the mapping object:
 
 .. code-block:: pycon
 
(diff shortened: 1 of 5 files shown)


docs/source/usage/rest_api.rst
```diff
@@ -85,12 +85,12 @@ To add a :ref:`mapping <mappings>` between previously-registered variation objec
    ...     json=payload
    ... )
 
-Mappings from an object can be retrieved via ``GET /variations/<vrs_id>/mappings/<mapping_type>``:
+Mappings from an object can be retrieved via ``GET /variations/<vrs_id>/mappings?mapping_type=<mapping_type>``:
```
theferrit32:
```suggestion
Mappings from an object can be retrieved via ``GET /object/<vrs_id>/mappings?mapping_type=<mapping_type>``:
```

docs/source/usage/rest_api.rst
theferrit32:
```suggestion
To add a :ref:`mapping <mappings>` between previously-registered variation objects issue a ``PUT`` request to ``/object/<vrs_id>/mappings``, where ``vrs_id`` is the ``source_id`` of the mapping object:
```

theferrit32 (approved):

PR #496, opened 2026-07-10
feat: retrieve catvars by way of mappings

### LINKED ISSUE #454: Build CatVar search endpoint
### Feature description

Users should be able to input a variant into a catvar search endpoint, and receive back all the CatVars registered in AnyVar of which the variant is a member.

### Use case

Being able to store CatVars is only helpful if I'm able to get them back out again.

### Acceptance Criteria

See Requirement #452 

### Proposed solution

A new HTTP endpoint that takes in a variant and outputs a list of CatVars. Perhaps we name this something like `GET /categorical_variations`?

### Alternatives considered

_No response_

### Implementation details

1. If a user searches a variant that's not a genomic variant, map it back down to the genomic version
2. Check the CatVar table for catvars w/ this genomic variant as the DefiningAlleleContexts
3. Liftover the genomic variant to the opposite build (i.e., if the searched variant was GRCh37, lift it over to GRCh38)
4. Check the CatVar table for catvars w/ this lifted-over genomic variant as the DefiningAlleleContexts

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

None

jsstevenson:
close #454 

### New endpoints

* `GET /categorical_variants?member_id=<VRS ID>`: given a registered variant ID, get all catvars of which it is a member
* `POST /categorical_variants`: using the same params for variant expressions that we employ elsewhere, translate the user input to an allele and retrieve catvars for which it is a member

### Refactor mapping services

Refactor liftover and projection services to be broken down into two kinds of external-facing functions (enabling projection of query variants w/o registering them):

1) "do the mappings" -- eg get the lifted-over allele, get the projected variant, etc
2) "do the mappings and register the new stuff" -- ie run the above function and store the result. 

In order to disentangle projection from immediate storage, I refactored a lot of the projection internal methods to progressively build an aggregate results class describing all successful/failed/skipped projection attempts. This provides a little more structure. Individual projection methods return an "outcome" (structured success/failure/reason for skipping), then all of them get combined into a big results object and returned to the API route. This also enables a little bit more transparency about stuff being skipped. I tried to tune down how much a `ProjectionError` was used to disrupt control flow so that partial progress doesn't get lost.

**Small caveat to the above**: I left a lot of the error/skip messages as-is even though they don't need as much detail given the additional structure. It would be pretty easy to go back and trim them down another time.

Define a new `protocols` module in the `mappings` subpackage to provide public interfaces to specific mapping functions without requiring import of those modules directly (i.e. because `coolseqtool` is currently an extra dependency). Would be good to implement a protocol for `agct` to ensure that liftover is truly optional too.

### Augment app-wide registration actions

Update registration to perform projection for a registered variant *and* its lifted-over equivalent (the genomic variants should both map to the same transcript, but it's maybe nice to ensure everything's filled in).

**Big caveat**: I realized fairly late that the core `register_variations` function (which, for some reason, lives in `src/anyvar/translate/`), wants to consume a raw variant expression, not a realized VRS object, so it can't be used when registering new catvars -- meaning we effectively need to define basic variant registration protocols multiple times (the `/vrs_variation` endpoint, which we are likely shuttering, also bypasses this function). Would be good to refactor this (I started trying to clean everything up and it quickly got big so I dropped it). In general, I think we are suffering from inconsistent structure about where and how "orchestrator"-type functions get defined.

## Behind the scenes

The idea is that when you submit a catvar member query, we 
1) attempt liftover/projection, giving us a list of allele IDs (not performed for the GET endpoint, since you're looking up a variant that's already been registered)
2) query the DB for all variations with mappings (transitive, undirected) from those alleles
3) return all catvars defined by that larger list of alleles

so, for example, given a canonical allele defined by genomic variant G that projects up to protein variant P:
1) user registers canonical allele. AnyVar registers defining allele G + liftover, transcript, protein projections
2) user queries by protein variant P. AnyVar first tries to project out from P (returning no additional variants) then queries for all variants that map to P (returning G) then looks up all catvars defined by the list `[P, G]` (returning the original canonical allele).


jsstevenson:
small follow up ticket: add additional tests to ensure proper functioning when certain things (liftover? projection?) are disabled or unavailable

tests/unit/mapping/test_projection.py
```diff
@@ -968,16 +889,19 @@ def test_project_transcript_variant_raises_missing_metadata_before_state_derivat
...
-        projection.ProjectionError,
-        match="Projection skipped: no CDS metadata for transcript NR_000001.1",
-    ):
-        projector._project_transcript_variant(variation, mocker.Mock(), "NR_000001.1")
+    # with pytest.raises(
+    #     projection.ProjectionError,
+    #     match=re.escape(
+    #         "Projection skipped: no CDS metadata for transcript NR_000001.1"
+    #     ),
+    # ):
```
jennifer-bowser: Whoops, looks like some commented-out code snuck through

jennifer-bowser (dismissed):
Looks good, except for one tiny bit of commented-out code I found!

jennifer-bowser (approved):

PR #502, opened 2026-07-30
ops: add alembic (issue #444)

### LINKED ISSUE #444: Add Alembic
Might need to add btreegist as part of this. If not, save that for another ticket. If so, make sure it doesn't mess with Snowflake databases.

jennifer-bowser:
closes #444 

This PR adds Alembic to AnyVar to manage database migrations. 

Just like before, AnyVar will automatically create all the required database tables if the DB hasn't already been initialized before. This is now done via Alembic when using Postgres, but is still done the original way for DuckDB. Alembic also handles installing the btreegist extension for Postgres instances.

I set it so that AnyVar will NOT automatically run DB migrations at start-up if the database has already been initialized but is using an outdated version of the schema. I figured since that could be a breaking change, we'd want to let users handle upgrades manually so they can do it when they're ready? Open to changing this though, I'd love thoughts from Reviewers here.

### Note
I didn't run tests locally for DuckDB since I don't have that installed, but everything seems to be passing in CI so I believe we're good to go?


alembic/env.py
```diff
@@ -0,0 +1,88 @@
...
+# from myapp import mymodel
+# target_metadata = mymodel.Base.metadata
+target_metadata = Base.metadata
+
+# other values from the config, defined by the needs of env.py,
+# can be acquired:
+# my_important_option = config.get_main_option("my_important_option")
+# ... etc.
+
+# Don't override config if it's already been set
```
jsstevenson:
```suggestion
config: Config = context.config

if config.config_file_name is not None:
    fileConfig(config.config_file_name)

target_metadata = Base.metadata
```

imo no need to include the boilerplate descriptions now that you've set everything up

alembic/env.py
```diff
@@ -0,0 +1,88 @@
...
+dotenv.load_dotenv()
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config: Config = context.config
+
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+if config.config_file_name is not None:
+    fileConfig(config.config_file_name)
```
jsstevenson, line 18: I don't think it's the end of the world, but it might be nicer to just reuse the existing logging setup function and keep all logging configs in one file rather than maintaining a separate config for migrations

alembic.ini
```diff
@@ -0,0 +1,149 @@
...
+# recursive_version_locations = false
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+# database URL.  This is consumed by the user-maintained env.py script only.
+# other means of configuring database URLs may be customized within the env.py
+# file.
+sqlalchemy.url = ""
```
jsstevenson: I know it's not directly conflicting as long as this is set to an empty string, but since we already have one path for setting this, I think we should just pull this one out of the template

alembic.ini
```diff
@@ -0,0 +1,149 @@
+# A generic, single database configuration.
```
jsstevenson, line 1: personal preference -- remove the unused configs rather than commenting them out; if we ever want to revise configs, the source of truth would be the alembic docs for whatever version we're on, not the boilerplate generated when we first commit the feature

alembic/env.py
```diff
@@ -0,0 +1,88 @@
+"""Alembic environment setup"""
+# ruff: noqa: INP001
```
jsstevenson: we should just turn INP001 off for the entire `alembic/` directory

alembic.ini
```diff
@@ -0,0 +1,149 @@
...
+# are written from script.py.mako
+# output_encoding = utf-8
+
+# database URL.  This is consumed by the user-maintained env.py script only.
+# other means of configuring database URLs may be customized within the env.py
+# file.
+sqlalchemy.url = ""
+
+
+[post_write_hooks]
```
jsstevenson, line 42: For convenience, it'd be nice to fire ruff-format and maybe ruff too

```
hooks = ruff,ruff-format
ruff.type = module
ruff.module = ruff
ruff.options = check --fix REVISION_SCRIPT_FILENAME

ruff-format.type = module
ruff-format.module = ruff
ruff-format.options = format REVISION_SCRIPT_FILENAME
``

alembic/env.py
jsstevenson: actually i think it's fine

alembic.ini
jsstevenson: upon further consideration, I think we actively do not want to set this here? need to keep poking around but I think we never want a blank url. and I'm also seeing a case where this gets interpreted as the string `'""'` which is truthy and prevents the url configuration from firing

jsstevenson (changes requested):
ok submitting an initial set of reviews. I'm trying to figure out if this duckdb issue can be worked around

alembic.ini
jennifer-bowser: Yeah no we _definitely_ never want to set this here. I think I meant to remove it instead of setting it to an empty string. Good catch!

jennifer-bowser:
> ok submitting an initial set of reviews. I'm trying to figure out if this duckdb issue can be worked around

@jsstevenson I addressed all your initial comments! I'd be _very_ down to add support for DuckDB here if we can though!!

jsstevenson (commented):
🚀 

we will solve the duckdb thing one day

